PR #9, opened 2022-02-09
Parameter Group Support

nmvk:
Memory DB Parameter Group support.

Commit contains e2e test framework required
for the MemoryDB.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,23 @@
+---
```
RedbackThomson, line 1: Please remove the `.github/ISSUE_TEMPLATE` directory. This is no longer needed now that we have a [`aws-controllers-k8s/.github` repository](https://github.com/aws-controllers-k8s/.github/). We do still need `.github/workflows`, though.

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,67 @@
+name: e2e-test
```
RedbackThomson, line 1: Please remove this workflow. All testing is done on Prow.

.github/workflows/unit-test-cover.yaml
```diff
@@ -0,0 +1,25 @@
+name: unit-test
```
RedbackThomson, line 1: Please remove this workflow as well.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+Issue #, if available:
```
RedbackThomson, line 1: Please remove this file.

apis/v1alpha1/generator.yaml
```diff
@@ -0,0 +1,43 @@
...
+    exceptions:
+      errors:
+        404:
+          code: ParameterGroupNotFoundFault
+      terminal_codes:
+        - ParameterGroupQuotaExceededFault
+        - ParameterGroupAlreadyExistsFault
+        - ServiceLinkedRoleNotFoundFault
+        - InvalidParameterGroupStateFault
+        - InvalidParameterValueException
```
RedbackThomson, line 12: Is this terminal? Is it possible to change the spec to undo this exception?

apis/v1alpha1/generator.yaml
```diff
@@ -0,0 +1,43 @@
...
+        - InvalidParameterCombinationException
```
RedbackThomson, line 13: Likewise with this exception.

apis/v1alpha1/generator.yaml
```diff
@@ -0,0 +1,43 @@
...
+        template_path: hooks/parametergroup/delta_post_compare.go.tpl
+
+
+ignore:
+  resource_names:
+    - Snapshot
+    - User
+    - ACL
+    - Cluster
+    - SubnetGroup
```
RedbackThomson, line 43: nit: Move to the top of the file

apis/v1alpha1/parameter_group.go
```diff
@@ -0,0 +1,89 @@
...
+type ParameterGroupSpec struct {
+	// An optional description of the parameter group.
+	Description *string `json:"description,omitempty"`
+	// The name of the parameter group family that the parameter group can be used
+	// with.
+	// +kubebuilder:validation:Required
+	Family *string `json:"family"`
+	// The name of the parameter group.
+	// +kubebuilder:validation:Required
+	ParameterGroupName *string `json:"parameterGroupName"`
```
RedbackThomson, line 37: Please rename this to just "name". See [the s3 controller](https://github.com/aws-controllers-k8s/s3-controller/blob/main/generator.yaml#L103-L113) for an example

pkg/resource/parameter_group/post_set_output.go
```diff
@@ -0,0 +1,90 @@
+package parameter_group
```
RedbackThomson, line 1: super nit: Add Amazon copyright headers to custom files. (I'm not sure if this is required tbh)

pkg/resource/parameter_group/post_set_output.go
```diff
@@ -0,0 +1,90 @@
...
+	ctx context.Context,
+	parameterGroupName *string,
+) ([]*svcapitypes.Parameter, error) {
+	parameters := []*svcapitypes.Parameter{}
+
+	input, err := rm.newDescribeParametersRequestPayload(parameterGroupName)
+	if err != nil {
+		return nil, err
+	}
+	response, respErr := rm.sdkapi.DescribeParametersWithContext(ctx, input)
```
RedbackThomson, line 49: It appears this method returns a [`NextToken`](https://docs.aws.amazon.com/sdk-for-go/api/service/memorydb/#DescribeParametersOutput) used for pagination. Check out how we handled that in the [IAM controller](https://github.com/aws-controllers-k8s/iam-controller/blob/main/pkg/resource/role/hooks.go#L217-L230).

pkg/resource/parameter_group/post_set_output.go
```diff
@@ -0,0 +1,90 @@
...
+	rm.metrics.RecordAPICall("READ_MANY", "DescribeParameters", respErr)
```
RedbackThomson, line 50:
```suggestion
	rm.metrics.RecordAPICall("GET", "DescribeParameters", respErr)
```

pkg/resource/parameter_group/update_pre_build.go
```diff
@@ -0,0 +1,73 @@
...
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+)
+
+func (rm *resourceManager) resetParameterGroup(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+) (updated *resource, err error) {
+	if (desired.ko.Spec.ParameterNameValues == nil || len(desired.ko.Spec.ParameterNameValues) == 0) && latest.ko.Spec.ParameterGroupName != nil {
```
RedbackThomson, line 14: `Spec.ParameterGroupName` is a required field, and is the primary identifier. I don't believe we need to check for `nil` anywhere for this particular property.

pkg/resource/parameter_group/update_pre_build.go
```diff
@@ -0,0 +1,73 @@
...
+
+// resetAllParameters resets cache parameters for given ParameterGroup in desired custom resource.
+func (rm *resourceManager) resetAllParameters(
+	ctx context.Context,
+	desired *resource,
+) (updated *resource, err error) {
+	input := &svcsdk.ResetParameterGroupInput{}
+	if desired.ko.Spec.ParameterGroupName != nil {
+		input.SetParameterGroupName(*desired.ko.Spec.ParameterGroupName)
+	}
```
RedbackThomson, line 30: If we don't set `ParameterGroupName` on the input, then this call would fail anyway. I think we can just assume it will always have a value.

pkg/resource/parameter_group/update_pre_build.go
```diff
@@ -0,0 +1,73 @@
...
+	input.SetAllParameters(true)
+
+	resp, err := rm.sdkapi.ResetParameterGroupWithContext(ctx, input)
+	rm.metrics.RecordAPICall("UPDATE", "ResetParameterGroup", err)
+	if err != nil {
+		rm.log.V(1).Info("Error during ResetParameterGroup", "error", err)
```
RedbackThomson, line 36: I think you can just return here. Logging can happen at a higher level, where it can be put into the condition message.

templates/hooks/parametergroup/delta_post_compare.go.tpl
```diff
@@ -0,0 +1,11 @@
+if delta.DifferentAt("Spec.ParameterGroupName") {
+    common.RemoveFromDelta(delta, "Spec.ParameterGroupName")
+}
+
+if delta.DifferentAt("Spec.ParameterGroupName") {
+    common.RemoveFromDelta(delta, "Spec.ParameterGroupName")
+}
```
RedbackThomson, line 7: This is duplicated. But regardless, what is the purpose of this?

test/e2e/bootstrap.yaml
```diff
@@ -0,0 +1 @@
+{}
```
RedbackThomson, line 1: I think you forgot to copy a `.gitignore` file in the `test/e2e/` directory

test/e2e/bootstrap_resources.py
```diff
@@ -0,0 +1,38 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
RedbackThomson, line 1: This is a very outdated version of test bootstrapping. It is no longer supported in `acktest`, so I would recommend you update it using the eks controller as your guide.

test/e2e/replacement_values.py
```diff
@@ -0,0 +1,18 @@
...
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+"""Stores the values used by each of the integration tests for replacing the
+RDS-specific test variables.
```
RedbackThomson, line 14: RDS -> `MemoryDB`

RedbackThomson (changes requested):
This looks like a well completed first resource for MemoryDB. I'm excited to see this controller gain traction. I've left some inline comments that I'd like you to have a look at.

In regards to the testing: I see you are using the elasticache controller's declarative testing framework. I would like to see all controller's move to declarative testing, however no-one has taken the time yet to standardise the framework and abstract it into the `acktest` Python package. The version you are using relies on outdated bootstrapping common code, which can be updated independently of the declarative stuff. However, I am worried about keeping this framework in line with the elasticache test directory.

pkg/resource/parameter_group/post_set_output.go
nmvk: API would return all the values and doesn't need pagination.

apis/v1alpha1/generator.yaml
nmvk: These are terminal errors, spec can undo all of them  except `ServiceLinkedRoleNotFoundFault` will remove it from the list.

templates/hooks/parametergroup/delta_post_compare.go.tpl
nmvk: Nice catch, Should have been `Spec.Family` only `ParameterNameValues` can be modified rest all are immutable

apis/v1alpha1/generator.yaml
nmvk: These are terminal errors, spec can undo all of them  except `ServiceLinkedRoleNotFoundFault` will remove it from the list.

go.mod
```diff
@@ -3,9 +3,11 @@ module github.com/aws-controllers-k8s/memorydb-controller
 go 1.17
 
 require (
-	github.com/aws-controllers-k8s/runtime v0.16.0
-	github.com/aws/aws-sdk-go v1.40.28
+	github.com/aws-controllers-k8s/runtime v0.16.4
```
vijtrip2, line 6: nit: latest runtime is v0.16.5 , but you can upgrade to it later as well.

pkg/common/delta.go
```diff
@@ -0,0 +1,42 @@
...
+			break
+		}
+	}
+
+	// if found, create a new slice and replace the original
+	if i != nil {
+		differences = append(differences[:*i], differences[*i+1:]...)
+		delta.Differences = differences
+	}
+}
```
vijtrip2, line 42:
```suggestion
	for i, diff := range differences {
		if diff.Path == subject {
	                differences = append(differences[:i], differences[i+1:]...)
		        delta.Differences = differences
		        return
		}
	}
}
```

pkg/resource/parameter_group/delta.go
```diff
@@ -0,0 +1,85 @@
...
+		common.RemoveFromDelta(delta, "Spec.Description")
+	}
+
+	if delta.DifferentAt("Spec.Name") {
+		common.RemoveFromDelta(delta, "Spec.Name")
+	}
+
+	if delta.DifferentAt("Spec.Family") {
+		common.RemoveFromDelta(delta, "Spec.Family")
+	}
```
vijtrip2, line 83: I am not sure about this. This will allow the reconciliation to be successful even when the desired and latest objects will not be same.

IMO, any time user updates these three fields, an error should be returned pointing them to the fact that these fields are immutable.

kumargauravsharma (commented):
Please check if ATTRIBUTION.md needs additional entries.

pkg/resource/parameter_group/delta.go
nmvk: For now I am ignoring this will work on `Advisory` condition later

kumargauravsharma (approved):

pkg/resource/parameter_group/post_set_output.go
RedbackThomson: > API would return all the values and doesn't need pagination.

But as I linked, the `DescribeParametersOutput` contains a `NextToken` field, with the comment:

> An optional argument to pass in case the total number of records exceeds
> the value of MaxResults. If nextToken is returned, there are more results
> available. The value of nextToken is a unique pagination token for each page.
> Make the call again using the returned token to retrieve the next page. Keep
> all other arguments unchanged.

Which implies the result is paginated and needs to be called multiple times to retrieve all the values?

pkg/resource/parameter_group/update_pre_build.go
```diff
@@ -0,0 +1,65 @@
+package parameter_group
```
RedbackThomson, line 1: nit: Needs Amazon header

RedbackThomson (changes requested):
Please add looping over the `DescribeParameters` including the `NextToken`. If we don't support the paginated responses now, it will be incredibly difficult to debug later.

Other than that it's good to go.

pkg/resource/parameter_group/post_set_output.go
```diff
@@ -0,0 +1,118 @@
...
+	if parameterGroupName != nil {
+		res.SetParameterGroupName(*parameterGroupName)
+	}
+
+	if nextToken != nil {
+		res.SetNextToken(*nextToken)
+	}
+
+	return res, nil
+}
```
A-Hilaly, line 118: nit: This function is not needed, you can build the request struct directly when calling `rm.sdkapi.DescribeParametersWithContext` in L64
e.g:

```go
rm.sdkapi.DescribeParametersWithContext(ctx, &svcsdk.DescribeParametersInput{
    ParameterGroupName: parameterGroupName,
    NextToken: nil,
})
```

pkg/resource/parameter_group/update_pre_build.go
```diff
@@ -0,0 +1,78 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package parameter_group
+
+import (
+	"context"
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
```
A-Hilaly, line 19: Please split and group the standard library, internal and external imports. https://github.com/golang/go/wiki/CodeReviewComments#imports

```go
import (
        "context"

	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
)
```

A-Hilaly (approved):
Looks good overall! left few comments and questions

templates/hooks/parametergroup/delta_post_compare.go.tpl
```diff
@@ -0,0 +1,11 @@
+if delta.DifferentAt("Spec.Description") {
```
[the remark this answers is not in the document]
nmvk: This is not required I removed this.

pkg/resource/parameter_group/update_pre_build.go
```diff
@@ -0,0 +1,78 @@
...
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+) (updated *resource, err error) {
+	if desired.ko.Spec.ParameterNameValues == nil || len(desired.ko.Spec.ParameterNameValues) == 0 {
+		// Reset all the parameters to default
+		return rm.resetAllParameters(ctx, desired)
+	}
+
+	return nil, nil
```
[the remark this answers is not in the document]
nmvk: No this has to `nil,nil` for invoking the other API.

RedbackThomson:
Let's ship it!
/lgtm

PR #11, opened 2022-02-28
Support for User, ACL, SubnetGroup

nmvk:
User
- Support for secret
- Re-queue logic to fix correct status
- E2E Tests
- Support for determining correct access string change.

ACL
- E2E Tests using User
- Re-queue logic to fix correct status
- E2E framework updated to create other MemoryDB resources in test.

SubnetGroup
- E2E Tests
- Determine correct subnets for E2E tests
- New bootstrap-able data to get correct memory db specific subnets.


vijtrip2 (commented):
Also please rebase with latest changes in main that got updated to ACK runtime v0.17.1

pkg/resource/acl/pre_build_request.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/pkg/errors"
+
+	"github.com/aws-controllers-k8s/runtime/pkg/requeue"
+)
+
+// validateACLNeedsUpdate this function's purpose is to requeue if the resource is currently unavailable
+func (rm *resourceManager) validateACLNeedsUpdate(
+	latest *resource,
+) (*resource, error) {
```
A-Hilaly, line 25: This function always returns a nil `*resource`, you can omit that and return only an error, or a boolean and an error if you wanna rename it `isValidACL`

pkg/resource/user/annotations.go
```diff
@@ -0,0 +1,67 @@
...
+) {
+	annotations := getAnnotationsFields(r, ko)
+	setLastRequestedAccessString(r, annotations)
+	ko.ObjectMeta.Annotations = annotations
+}
+
+// getAnnotationsFields return the annotations map that would be used to set the fields
+func getAnnotationsFields(
+	r *resource,
+	ko *svcapitypes.User) map[string]string {
```
A-Hilaly, line 42:
```suggestion
	ko *svcapitypes.User,
) map[string]string {
```

pkg/resource/user/annotations.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	desiredAnnotations := r.ko.ObjectMeta.GetAnnotations()
+	annotations := make(map[string]string)
+	for k, v := range desiredAnnotations {
+		annotations[k] = v
+	}
+
+	ko.ObjectMeta.Annotations = annotations
+	return annotations
```
A-Hilaly, line 55: Why do you need to copy annotations here? maybe `return r.ko.ObjectMeta.Annotations` is enough?

test/e2e/declarative_test_fwk/model.py
```diff
@@ -146,13 +146,18 @@ def __repr__(self) -> str:
...
+    steps: List[StepDict]
+
+
 # fields in a test scenario
 class ScenarioDict(TypedDict, total=False):
     id: str
     description: str
     marks: List[str]
-    resource: ResourceDict
-    steps: List[StepDict]
```
kumargauravsharma, line 155: Resource can be specified at each step as create, patch, delete allow specifying ResourceDict attributes at each step.

apis/v1alpha1/generator.yaml
```diff
@@ -1,11 +1,126 @@
...
+            SubnetGroupName: Name
+  ACL:
+    exceptions:
+      errors:
+        404:
+          code: ACLNotFoundFault
+      terminal_codes:
+        - ACLAlreadyExistsFault
+        - DefaultUserRequired
+        - UserNotFoundFault
```
vijtrip2, line 49: nit: recommended to remove NotFound errors from terminal code because referencing resources it is usual to see NotFound error while referenced resource is being created

templates/hooks/subnetgroup/sdk_create_post_set_output.go.tpl
```diff
@@ -0,0 +1,16 @@
+var resourceSyncedCondition *ackv1alpha1.Condition = nil
```
vijtrip2, line 1: You should not be needing Custom hook for setting Synced condition. If there are no errors and this condition will set to true by default at the end of reconciliation loop. If you want to control it then you can use SyncedConfig inside generator.yaml.

vijtrip2 (approved):
looks good. two minor comments.

PR #14, opened 2022-03-17
Snapshot Resource changes for Ack

AnilYalavarthi:
Description of changes: Adding Snapshot Resource for MemoryDB


apis/v1alpha1/snapshot.go
```diff
@@ -0,0 +1,91 @@
...
+// Represents a copy of an entire cluster as of the time when the snapshot was
+// taken.
+type SnapshotSpec struct {
+	// The snapshot is created from this cluster.
+	ClusterName *string `json:"clusterName,omitempty"`
+	// The ID of the KMS key used to encrypt the snapshot.
+	KMSKeyID *string `json:"kmsKeyID,omitempty"`
+	// A name for the snapshot being created.
+	// +kubebuilder:validation:Required
+	SnapshotName *string `json:"snapshotName"`
```
RedbackThomson, line 34: Can you please rename this just `Name`? Use the [S3 controller](https://github.com/aws-controllers-k8s/s3-controller/blob/main/generator.yaml#L103-L113) for reference

apis/v1alpha1/generator.yaml
```diff
@@ -1,11 +1,51 @@
...
+      errors:
+        404:
+          code: SnapshotNotFoundFault
+      terminal_codes:
+        -  InvalidParameterCombinationException
+        -  InvalidSnapshotStateFault
+        -  SnapshotAlreadyExistsFault
+        -  SnapshotQuotaExceededFault
+        -  TagQuotaPerResourceExceeded
+        -  ClusterNotFoundFault
```
RedbackThomson, line 22: Please remove this from terminal. We want the controller to retry until the Cluster is found!

apis/v1alpha1/generator.yaml
```diff
@@ -1,11 +1,51 @@
...
+        -  InvalidClusterStateFault
```
RedbackThomson, line 23: Maybe we should remove this one as well, if there is a chance the cluster will recover into a good state.

apis/v1alpha1/generator.yaml
```diff
@@ -1,11 +1,51 @@
...
+        -  InvalidParameterValueException
+        -  InvalidParameter
+    fields:
+      SourceSnapshotName:
+        from:
+          operation: CopySnapshot
+          path: SourceSnapshotName
+      ClusterName:
```
RedbackThomson, line 31: Is this field not automatically generated? Why do we have to add it to the list of fields?

apis/v1alpha1/snapshot.go
```diff
@@ -0,0 +1,91 @@
...
+	// the various terminal states of the CR and its backend AWS service API
+	// resource
+	// +kubebuilder:validation:Optional
+	Conditions []*ackv1alpha1.Condition `json:"conditions"`
+	// The configuration of the cluster from which the snapshot was taken
+	// +kubebuilder:validation:Optional
+	ClusterConfiguration *ClusterConfiguration `json:"clusterConfiguration,omitempty"`
+	// The name of the snapshot
+	// +kubebuilder:validation:Optional
+	Name *string `json:"name,omitempty"`
```
RedbackThomson, line 60: This field should disappear if we rename the spec field - we don't want it duplicated across spec and status.

pkg/resource/snapshot/copy_snapshot.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (rm *resourceManager) customCreateSnapshot(
+	ctx context.Context,
+	r *resource,
+) (*resource, error) {
+	if r.ko.Spec.SourceSnapshotName != nil {
```
RedbackThomson, line 29: nit: short circuit rather than nesting in an if statement

```suggestion
	if r.ko.Spec.SourceSnapshotName == nil {
	  return nil, nil
	}
```

pkg/resource/snapshot/copy_snapshot.go
```diff
@@ -0,0 +1,185 @@
...
+		if r.ko.Spec.ClusterName != nil {
+			return nil, awserr.New("InvalidParameterCombinationException", "Cannot specify CacheClusteId"+
```
RedbackThomson, line 31: Please just return a regular go error using `errors.New()`. We reserve `awserr` for errors returned by the APIs.

pkg/resource/snapshot/custom_set_output.go
```diff
@@ -0,0 +1,90 @@
...
+package snapshot
+
+import (
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	"github.com/aws/aws-sdk-go/service/memorydb"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func (rm *resourceManager) CustomDescribeSnapshotSetOutput(
```
RedbackThomson, line 23: Any reason why these methods are exported? Unless we are unit testing them elsewhere, I'd recommend we leave them un-exported (lowercase first letter)

pkg/resource/snapshot/custom_set_output.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if resourceSyncedCondition == nil {
+		resourceSyncedCondition = &ackv1alpha1.Condition{
+			Type:   ackv1alpha1.ConditionTypeResourceSynced,
+			Status: syncConditionStatus,
+		}
+		ko.Status.Conditions = append(ko.Status.Conditions, resourceSyncedCondition)
+	} else {
+		resourceSyncedCondition.Status = syncConditionStatus
+	}
```
RedbackThomson, line 89: We can actually do all of this in the generator.yaml file, now! Check out our test case for this: https://github.com/aws-controllers-k8s/code-generator/blob/d2ebb1c8a5e181814acc18a0cfc165e449f98504/pkg/testdata/models/apis/lambda/0000-00-00/generator.yaml#L14-L19

pkg/resource/snapshot/utility.go
```diff
@@ -0,0 +1,151 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	condMsgCurrentlyDeleting string = "snapshot currently being deleted."
+)
+
+var (
+	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
+		errors.New("Delete is in progress."),
```
RedbackThomson, line 18: nit: lowercase error messages and remove trailing period
```suggestion
		errors.New("delete is in progress"),
```

pkg/resource/snapshot/utility.go
```diff
@@ -0,0 +1,151 @@
...
+	"errors"
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	condMsgCurrentlyDeleting string = "snapshot currently being deleted."
```
RedbackThomson, line 13:
```suggestion
	condMsgCurrentlyDeleting string = "snapshot currently being deleted"
```

pkg/resource/snapshot/utility.go
```diff
@@ -0,0 +1,151 @@
...
+	)
+)
+
+// isDeleting returns true if supplied snapshot resource state is 'deleting'
+func isDeleting(r *resource) bool {
+	if r == nil || r.ko.Status.Status == nil {
+		return false
+	}
+	status := *r.ko.Status.Status
+	return status == "deleting"
```
RedbackThomson, line 29: Weird, usually you can find these string constants in the Go SDK - https://docs.aws.amazon.com/sdk-for-go/api/service/memorydb/#pkg-constants

But I can't find anything for `"deleting"`. Would you mind making it a constant at the top of this package, though?

templates/hooks/snapshot/sdk_create_post_set_output.go.tpl
```diff
@@ -0,0 +1,5 @@
+// custom set output from response
```
RedbackThomson, line 1: Please indent all hook templates with one tab

pkg/resource/snapshot/custom_set_output.go
```diff
@@ -0,0 +1,90 @@
...
+) (*svcapitypes.Snapshot, error) {
+	rm.customSetOutput(resp.Snapshot, ko)
+	return ko, nil
+}
+
+func (rm *resourceManager) CustomCopySnapshotSetOutput(
+	resp *memorydb.CopySnapshotOutput,
+	ko *svcapitypes.Snapshot,
+) *svcapitypes.Snapshot {
+	rm.customSetOutput(resp.Snapshot, ko)
```
RedbackThomson, line 47: Why is this necessary for create? (Ignoring conditions) Does `KMSKeyID` not get populated correctly from the response?

pkg/resource/snapshot/copy_snapshot.go
```diff
@@ -0,0 +1,185 @@
...
+import (
+	"context"
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (rm *resourceManager) customCreateSnapshot(
```
RedbackThomson, line 25: Could you please rename this to something that denotes it's attempting to call CopySnapshot? Perhaps `customTryCopySnapshot`?

templates/hooks/snapshot/sdk_delete_post_request.go.tpl
```diff
@@ -0,0 +1,17 @@
...
+	// Setting resource synced condition to false will trigger a requeue of
+	// the resource.
+	ackcondition.SetSynced(
+		rp,
+		corev1.ConditionFalse,
+		&condMsgCurrentlyDeleting,
+		nil,
+	)
+	// Need to return a requeue error here, otherwise:
+	// - reconciler.deleteResource() marks the resource unmanaged
```
RedbackThomson, line 12: What's wrong with marking it unmanaged? Do you want to keep reconciling it until it actually 404s? I think some controllers just let things disappear from k8s once they have confirmed it's in "deleting" status.

templates/hooks/snapshot/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,2 @@
+//Not used currently but whenever SDKUpdate() is generated for non existing update() operation on resource this will be added then as override
+return latest, nil
```
RedbackThomson, line 2: You're probably safe to just delete this for now. If update isn't supported yet, better to return the not implemented error.

test/e2e/bootstrappable/snapshots.py
```diff
@@ -0,0 +1,60 @@
...
+from dataclasses import dataclass
+
+import boto3
+import time
+from acktest.bootstrapping import Bootstrappable
+from acktest.resources import random_suffix_name
+
+
+@dataclass
+class Snapshots(Bootstrappable):
```
RedbackThomson, line 23: Great job writing a `Bootstrappable` type! Really happy to see it being used.

This `Bootstrappable` is actually bootstrapping a `MemoryDB` cluster - so should probably be called `Cluster`

```suggestion
class Cluster(Bootstrappable):
```

test/e2e/bootstrappable/snapshots.py
```diff
@@ -0,0 +1,60 @@
...
+        available_status = "Available"
+        while True:
+            clusters = mdb.describe_clusters(ClusterName=self.snapshotClusterName)
+            cluster = clusters['Clusters'][0]
+            if cluster.get("Status").casefold() == available_status.casefold():
+                return True
+            if time.time() > timeout:
+                break
+            time.sleep(60)
+        raise ValueError('cluster not created within expected time')
```
RedbackThomson, line 42: Ouch there are no waiters supported in boto3 for memorydb yet

test/e2e/bootstrappable/snapshots.py
```diff
@@ -0,0 +1,60 @@
...
+import boto3
+import time
+from acktest.bootstrapping import Bootstrappable
+from acktest.resources import random_suffix_name
+
+
+@dataclass
+class Snapshots(Bootstrappable):
+    # Output
+    snapshotClusterName: str
```
RedbackThomson, line 25: Outputs need to be instantiated with an initialised value:
```suggestion
    cluster_name str = field(init=False)
```

test/e2e/bootstrap_resources.py
```diff
@@ -18,12 +18,13 @@
...
 from e2e import bootstrap_directory
+from e2e.bootstrappable.snapshots import Snapshots
 
 @dataclass
 class BootstrapResources(Resources):
 
-    def replacement_dict(self):
-        return {}
+    Snapshot1: Snapshots
+    Snapshot2: Snapshots
```
RedbackThomson, line 27: Is there a reason we need two clusters? Are we able to run both tests on a single cluster? It would certainly save us some time when bootstrapping.

RedbackThomson (changes requested):
Anil, this is an amazing job! There has clearly been a lot of work put into this PR and it looks great. You definitely chose to tackle a harder API, given the complexity in the `CopySnapshot` option.

I've left you some comments inline regarding some ways we can save on custom code, and to keep things more inline with other ACK controllers.

apis/v1alpha1/generator.yaml
nmvk: This is part of different API (copy-snapshot) so we would need to add it

test/e2e/replacement_values.py
```diff
@@ -23,4 +23,6 @@
     "TOPIC1": get_bootstrap_resources().Topic1.topic_arn,
     "TOPIC2": get_bootstrap_resources().Topic2.topic_arn,
     "KMSKEY": get_bootstrap_resources().KMSKey.key
+    "SNAPSHOT_CLUSTER_NAME1": get_bootstrap_resources().Cluster1.clusterName,
```
kumargauravsharma, line 26: It seems e2e test failed due to missing comma after the KMSKEY statement.
```
tests/test_scenarios.py:29: in <module>
    from e2e.replacement_values import REPLACEMENT_VALUES
E     File "/memorydb-controller/tests/e2e/replacement_values.py", line 26
E       "SNAPSHOT_CLUSTER_NAME1": get_bootstrap_resources().Cluster1.clusterName,
E       ^
E   SyntaxError: invalid syntax
```

test/e2e/replacement_values.py
```diff
@@ -22,5 +22,7 @@
     "SUBNET2": get_bootstrap_resources().Subnets.subnets[1],
     "TOPIC1": get_bootstrap_resources().Topic1.topic_arn,
     "TOPIC2": get_bootstrap_resources().Topic2.topic_arn,
-    "KMSKEY": get_bootstrap_resources().KMSKey.key
+    "KMSKEY": get_bootstrap_resources().KMSKey.key,
+    "SNAPSHOT_CLUSTER_NAME1": get_bootstrap_resources().Cluster1.clusterName,
+    "SNAPSHOT_CLUSTER_NAME2": get_bootstrap_resources().Cluster1.clusterName
```
nmvk, line 27: Should one of them be Cluster 2? We cannot take two snapshot from same cluster at same time.

kumargauravsharma:
e2e tests have passed, merging the PR.

kumargauravsharma (approved):

PR #17, opened 2022-03-30
MemoryDB Cluster resource support

nmvk:
Support for MDB cluster resource.

- Ensure that we invoke update API if there delta
- Scale up before scale in, Scale out before scale down
- Use annotation to rollback in case of async failures
- E2E tests to cover Create/Update/Rollback
- TerminalConditions for User errors
- Wait for delete to be completed
- Perform engineversion upgrade and scale up/down whenever possible


apis/v1alpha1/generator.yaml
```diff
@@ -1,8 +1,74 @@
...
+      errors:
+        404:
+          code: ClusterNotFoundFault
+      terminal_codes:
+        - ClusterAlreadyExistsFault
+        - ClusterQuotaForCustomerExceededFault
+        - NodeQuotaForClusterExceededFault
+        - NodeQuotaForCustomerExceededFault
+        - InsufficientClusterCapacityFault
+        - InvalidVPCNetworkStateFault
```
RedbackThomson, line 16: Not a terminal code: VPC could recover

apis/v1alpha1/generator.yaml
```diff
@@ -1,8 +1,74 @@
...
+        - ShardsPerClusterQuotaExceededFault
+        - InvalidParameterValueException
+        - InvalidParameterCombinationException
+        - InvalidCredentialsException
```
RedbackThomson, line 20: Not a terminal code: credentials may change permissions

apis/v1alpha1/generator.yaml
```diff
@@ -1,8 +1,74 @@
...
+        - TagQuotaPerResourceExceeded
+        - InvalidACLStateFault
+        - InvalidNodeStateFault
+        - InvalidKMSKeyFault
```
RedbackThomson, line 24: Not a terminal code: these resources may change state (existence)

pkg/resource/cluster/annotations.go
```diff
@@ -0,0 +1,82 @@
...
+	numShards *int64,
+	ko *svcapitypes.Cluster,
+) {
+	if ko.ObjectMeta.Annotations == nil {
+		ko.ObjectMeta.Annotations = make(map[string]string)
+	}
+
+	annotations := ko.ObjectMeta.Annotations
+
+	if numShards != nil {
```
RedbackThomson, line 45: nit: For each of these `set*Annotation` methods, change the input from `*int64` to `int64` and do nil-validation before calling these methods

pkg/resource/cluster/hooks.go
```diff
@@ -0,0 +1,54 @@
...
+	)
+)
+
+// isDeleting returns true if supplied cluster resource state is 'deleting'
+func isDeleting(r *resource) bool {
+	if r == nil || r.ko.Status.Status == nil {
+		return false
+	}
+	status := *r.ko.Status.Status
+	return status == "deleting"
```
RedbackThomson, line 44: Ugh why does the SDK not have this as a constant? Weird inconsistency. Would you mind creating a const at the top of this file and using that, at least? (Same for updating below)

pkg/resource/cluster/pre_build_request.go
```diff
@@ -0,0 +1,74 @@
...
+// validateClusterNeedsUpdate this function's purpose is to requeue if the resource is currently unavailable and
+// to validate if resource update is required.
+func (rm *resourceManager) validateClusterNeedsUpdate(
+	desired *resource,
+	latest *resource,
+	delta *ackcompare.Delta,
+) (*resource, error) {
+	// requeue if necessary
+	latestStatus := latest.ko.Status.Status
+	if latestStatus == nil || *latestStatus != "available" {
```
RedbackThomson, line 34: Same goes for these status - please make them into constants

pkg/resource/cluster/pre_build_request.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, requeue.NeededAfter(
+			errors.New("Cluster cannot be updated as its status is not 'active'."),
```
RedbackThomson, line 36: nit: errors should start with a lowercase (as is the convention in go) and should not end with a period.

Also this says `is not 'active'` whereas you're checking for `available` above. You could use `fmt.Errorf()` to put the constant you create above into this error string

pkg/resource/cluster/pre_build_request.go
```diff
@@ -0,0 +1,74 @@
...
+			requeue.DefaultRequeueAfterDuration)
+	}
+
+	// Set terminal condition when cluster is in create-failed state
+	if latestStatus != nil || *latestStatus == "create-failed" {
+		return nil, awserr.New("InvalidNodeStateFault", "Cluster is in create-failed state, cannot be updated.", nil)
```
RedbackThomson, line 42: nit: Please don't create an `awserr` type. These are reserved from errors returned by the AWS API

pkg/resource/cluster/update_post_build.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func (rm *resourceManager) newMemoryDBClusterUploadPayload(
+	desired *resource,
+	latest *resource,
+	delta *ackcompare.Delta,
+) *svcsdk.UpdateClusterInput {
+	res := &svcsdk.UpdateClusterInput{}
+
+	if delta.DifferentAt("Spec.AclName") && desired.ko.Spec.ACLName != nil {
```
RedbackThomson, line 28: This should be `Spec.ACLName`?

pkg/resource/cluster/update_post_build.go
```diff
@@ -0,0 +1,123 @@
...
+	// Determine if we are trying to scale up an instance
+	scaleUpUpdate := false
+	if delta.DifferentAt("Spec.NodeType") && desired.ko.Spec.NodeType != nil {
+		for _, instance := range desired.ko.Status.AllowedScaleUpNodeTypes {
+			if *instance == *desired.ko.Spec.NodeType {
+				scaleUpUpdate = true
+				break
+			}
+		}
+	}
```
RedbackThomson, line 78: This little complexity in the API kind of shows why only adding deltas to the Update types is non-trivial. We can only set some fields conditional on others. We can't exactly add a delta condition to every field, and copy the field contents over 1:1.

templates/hooks/cluster/sdk_read_many_post_set_output.go.tpl
```diff
@@ -0,0 +1,26 @@
+	if resp.Clusters[0].NumberOfShards != nil {
```
RedbackThomson, line 1: nit: early exit for if `len(resp.Clusters) < 1`, then you can just set `cluster := resp.Clusters[0]`

test/e2e/bootstrappable/topics.py
```diff
@@ -0,0 +1,44 @@
...
+
+from dataclasses import dataclass, field
+
+from acktest.bootstrapping import Bootstrappable
+from acktest.resources import random_suffix_name
+import boto3
+
+
+@dataclass
+class Topics(Bootstrappable):
```
RedbackThomson, line 23: Probably better to just make this `class Topic` and add 2 of them to your `BootstrapResources` object? `topic1` and `topic2` really don't have much meaning, and there may be a case where you want 3 later?

RedbackThomson (changes requested):
I haven't taken a chance to look at the tests yet, but really happy with the hook code you've added. It seems like there is a lot of complexity in these APIs and you've done a good job of separating it out into manageable chunks. I've left some comments inline regarding common code practices and some touchups. I'll take a look at tests and a little further into how we could re-use some of this code on the next review.

pkg/resource/cluster/update_post_build.go
nmvk: I agree, we it is not trivial to generate the update payload logic lot of complexity involved.

nmvk: I agree, we it is not trivial to generate the update payload logic lot of complexity involved.

pkg/resource/cluster/hooks.go
```diff
@@ -0,0 +1,61 @@
...
+		errors.New("update is in progress"),
+		ackrequeue.DefaultRequeueAfterDuration,
+	)
+)
+
+const (
+	available    = "available"
+	deleting     = "deleting"
+	updating     = "updating"
+	createFailed = "create-failed"
```
RedbackThomson, line 42: nit: Please append const names with `Status`. eg. `StatusAvailable`. This is to be consistent with other controllers that do the same (and provides some context around what they're describing)

pkg/resource/cluster/post_set_output.go
```diff
@@ -0,0 +1,58 @@
...
+) {
+	if *ko.Status.Status != available {
+		return
+	}
+
+	response, respErr := rm.sdkapi.ListAllowedNodeTypeUpdatesWithContext(ctx, &svcsdk.ListAllowedNodeTypeUpdatesInput{
+		ClusterName: ko.Spec.Name,
+	})
+	rm.metrics.RecordAPICall("GET", "ListAllowedNodeTypeUpdates", respErr)
+	// Ignore the error since the response from this API is used for information purpose only
```
RedbackThomson, line 53: If it throws an error, we won't be updating the spec or status properly, though? This could lead to a faulty delta and therefore a faulty update mechanism. I think I'd rather we just catch it and requeue the resource to be safe.

pkg/resource/cluster/pre_build_request.go
```diff
@@ -0,0 +1,74 @@
...
+func (rm *resourceManager) validateClusterNeedsUpdate(
+	desired *resource,
+	latest *resource,
+	delta *ackcompare.Delta,
+) (*resource, error) {
+	// requeue if necessary
+	latestStatus := latest.ko.Status.Status
+	if latestStatus == nil || *latestStatus != available {
+		return nil, requeue.NeededAfter(
+			errors.New("cluster cannot be updated as its status is not 'available'"),
```
RedbackThomson, line 36: Please use `fmt.Errorf`:
```suggestion
			fmt.Errorf("cluster cannot be updated as its status is not '%s'", StatusAvailable),
```

pkg/resource/cluster/pre_build_request.go
```diff
@@ -0,0 +1,74 @@
...
+			requeue.DefaultRequeueAfterDuration)
+	}
+
+	// Set terminal condition when cluster is in create-failed state
+	if *latestStatus == createFailed {
+		return nil, ackerr.NewTerminalError(errors.New("cluster is in create-failed state, cannot be updated"))
```
RedbackThomson, line 42: Likewise use `fmt.Errorf`

test/e2e/bootstrappable/topic.py
```diff
@@ -0,0 +1,40 @@
...
+import boto3
+
+
+@dataclass
+class Topic(Bootstrappable):
+    # Output
+    topic_arn: str = field(init=False)
+
+    def bootstrap(self):
+        """Create a two SNS topic.
```
RedbackThomson, line 28: Oops

RedbackThomson (changes requested):
Just a few small nits and then I'm happy.

I do have a bit of a problem with these declarative tests in that we are only checking the fields of the K8s resource. We never describe the state of the actual MemoryDB cluster. It feels like it's kind of defeating the purpose of an e2e test, if we don't test what we're integrating with. This probably needs to be figured out before we can push declarative tests through to all controllers

RedbackThomson:
Thank you!
/lgtm

PR #31, opened 2022-10-04
Solve the problem that customers cannot update tags after creation.

kyriechen96:
Issue #, if available:

Description of changes:
Deletes previous tags function which didn't work, and adds a new tags function that customer can update tags.


kyriechen96:
Wrong change

PR #32, opened 2022-10-26
Support Tagging API for parameter group

kyriechen96:
Support Tagging API for parameter group, so customers could update tags for their parameter groups after creation.

Issue #, if available:

Description of changes: Modifies existing hooks and templates. Adds new hooks and templates to support tagging API.


pkg/resource/parameter_group/read_post_set_output.go
```diff
@@ -0,0 +1,31 @@
...
+	ctx context.Context,
+	resourceARN string,
+) ([]*svcapitypes.Tag, error) {
+	resp, err := rm.sdkapi.ListTagsWithContext(
+		ctx,
+		&svcsdk.ListTagsInput{
+			ResourceArn: &resourceARN,
+		},
+	)
+	rm.metrics.RecordAPICall("GET", "ListTagsForResource", err)
```
nmvk, line 19: API name is `ListTags`

pkg/resource/parameter_group/update_post_set_update.go
```diff
@@ -0,0 +1,95 @@
...
+	if len(toDelete) > 0 {
+		rlog.Debug("removing tags from parameter group", "tags", toDelete)
+		_, err = rm.sdkapi.UntagResourceWithContext(
+			ctx,
+			&svcsdk.UntagResourceInput{
+				ResourceArn: arn,
+				TagKeys:     toDelete,
+			},
+		)
+		rm.metrics.RecordAPICall("UPDATE", "RemoveTagsFromResource", err)
```
nmvk, line 35: API Name is `UnTagResource`

templates/hooks/parametergroup/sdk_update_pre_build_request.go.tpl
```diff
@@ -4,4 +4,6 @@ if delta.DifferentAt("Spec.ParameterNameValues") {
     if ko != nil || err != nil {
         return ko, err
     }
-}
\ No newline at end of file
+}
+
+
```
nmvk, line 9: nit: Remove the extra lines

templates/hooks/parametergroup/sdk_update_post_set_output.go.tpl
```diff
@@ -0,0 +1,13 @@
...
+if err != nil {
+    return nil, err
+}
+
+if delta.DifferentAt("Spec.Tags") {
+    err = rm.updateTags(ctx, ko, desired, latest);
+}
+
+if err != nil {
+    return nil, err
```
nmvk, line 12: Why we cannot do this in `sdk_update_pre_build`?

pkg/resource/parameter_group/update_post_set_update.go
```diff
@@ -0,0 +1,95 @@
...
+	if len(toAdd) > 0 {
+		rlog.Debug("adding tags to parameter group", "tags", toAdd)
+		_, err = rm.sdkapi.TagResourceWithContext(
+			ctx,
+			&svcsdk.TagResourceInput{
+				ResourceArn: arn,
+				Tags:        sdkTagsFromResourceTags(toAdd),
+			},
+		)
+		rm.metrics.RecordAPICall("UPDATE", "AddTagsToResource", err)
```
nmvk, line 50: Update API name here too

nmvk (changes requested):

pkg/resource/parameter_group/update_post_set_update.go
kyriechen96: Done.

templates/hooks/parametergroup/sdk_update_post_set_output.go.tpl
kyriechen96: Yes, I just changed it.

templates/hooks/parametergroup/sdk_update_pre_build_request.go.tpl
kyriechen96: Got it.

pkg/resource/parameter_group/update_post_set_update.go
kyriechen96: Done.

pkg/resource/parameter_group/read_post_set_output.go
kyriechen96: Done.

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2022-04-15T17:21:57Z"
-  build_hash: 50c64871bcaf88b9ee200eb8d6b8245fa8f675eb
-  go_version: go1.17.5
-  version: v0.18.4
-api_directory_checksum: 6972849b7a809751f78bbafa31eeb725ed5bc126
+  build_date: "2022-10-26T20:52:18Z"
+  build_hash: 5ee0ac052c54f008dff50f6f5ebb73f2cf3a0bd7
+  go_version: go1.19
+  version: 5ee0ac0
```
A-Hilaly, line 5: We need to keep the `ack-generate` and `runtime` versions equal (make sure to fetch the code-generator tags `git fetch --tags --all` before the generation)

A-Hilaly: Ideally we should generate the controller using version `v0.20.1` after bumping the runtime version to `v0.20.1` (PR #30)

pkg/resource/parameter_group/post_set_output.go
```diff
@@ -15,7 +15,6 @@ package parameter_group
 
 import (
 	"context"
-
```
A-Hilaly, line 18: Please keep the imports organized using the Go best practices mentioned in https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/resource/parameter_group/read_post_set_output.go
```diff
@@ -0,0 +1,31 @@
+package parameter_group
+
+import (
+	"context"
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
```
A-Hilaly, line 6: Same comment regarding the order and the separation of the imports

pkg/resource/parameter_group/sdk.go
```diff
@@ -18,13 +18,15 @@ package parameter_group
...
 	"errors"
+	"fmt"
 	"reflect"
 	"strings"
 
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
 	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
+	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
```
A-Hilaly, line 29: This is a good example of how to seperate/order import paths

pkg/resource/parameter_group/sdk.go
```diff
@@ -112,6 +118,15 @@ func (rm *resourceManager) sdkFind(
...
 	rm.setStatusDefaults(ko)
+	if ko.Status.ACKResourceMetadata != nil && ko.Status.ACKResourceMetadata.ARN != nil {
+		resourceARN := (*string)(ko.Status.ACKResourceMetadata.ARN)
+		tags, err := rm.getTags(ctx, *resourceARN)
+		if err != nil {
+			return nil, err
+		}
+		ko.Spec.Tags = tags
+	}
+
```
A-Hilaly, line 129: There is no need to perform a nil check operation on `ko.Status.ACKResourceMetadata` and `ko.Status.ACKResourceMetadata.ARN` here. If the execution makes it to this block it means that a Parameter group exists and you'll have an ARN 100%. Check L116-L118

pkg/resource/parameter_group/sdk.go
```diff
@@ -247,14 +264,25 @@ func (rm *resourceManager) sdkUpdate(
...
 		}
 	}
+
+	if delta.DifferentAt("Spec.Tags") {
+		err = rm.updateTags(ctx, desired, latest)
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
A-Hilaly, line 284: The error check here should be right after L279. Otherwise, if there is no difference at `Spec.Tag` it's gonna evaluate the error declared in L264

pkg/resource/parameter_group/update_pre_build_request.go
```diff
@@ -15,6 +15,8 @@ package parameter_group
 
 import (
 	"context"
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
```
A-Hilaly, line 19: Same comment on import paths

pkg/resource/parameter_group/update_pre_build_request.go
```diff
@@ -15,6 +15,8 @@ package parameter_group
```
A-Hilaly, line 15: Why was this file renamed to `update_pre_build_request.go`? maybe better to just name it `hooks.go`?

pkg/resource/parameter_group/update_pre_build_request.go
```diff
@@ -77,3 +79,88 @@ func (rm *resourceManager) resetAllParameters(
...
+func computeTagsDelta(
+	desired []*svcapitypes.Tag,
+	latest []*svcapitypes.Tag,
+) (added []*svcapitypes.Tag, removed []*string) {
+	toDelete := []*string{}
+	toAdd := []*svcapitypes.Tag{}
+
+	desiredTags := map[string]string{}
+	for _, tag := range desired {
+		desiredTags[*tag.Key] = *tag.Value
```
A-Hilaly, line 140: In case the user provides a desired tag key but no value, the code is going to panic here

templates/hooks/parametergroup/sdk_update_pre_build_request.go.tpl
```diff
@@ -4,4 +4,12 @@ if delta.DifferentAt("Spec.ParameterNameValues") {
...
\ No newline at end of file
+}
+
+if delta.DifferentAt("Spec.Tags") {
+    err = rm.updateTags(ctx, desired, latest);
+}
+
+if err != nil {
+    return nil, err
+}
```
A-Hilaly, line 15:
```suggestion
if delta.DifferentAt("Spec.Tags") {
    err = rm.updateTags(ctx, desired, latest);
    if err != nil {
        return nil, err
    }
}
```

pkg/resource/parameter_group/update_pre_build_request.go
```diff
@@ -77,3 +79,88 @@ func (rm *resourceManager) resetAllParameters(
 	}
 	return &resource{ko}, nil
 }
+
+func (rm *resourceManager) updateTags(
```
A-Hilaly, line 83: it would nice having a comment describing what this function is doing.

A-Hilaly (changes requested):
Nice, thank you @kyriechen96!
I left few comments and suggestions bellow.

pkg/resource/parameter_group/sdk.go
kyriechen96: Thank you for the example! Just a quick question. For a group of imports, shouldn't them be ordered alphabetically? Like putting ackcompare before ackv1alpha1?

A-Hilaly: Good question! It's organized alphabetically using the import path not the import alias :)
For example in `ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"`, `ackcondition` is the import alias and `"github.com/aws-controllers-k8s/runtime/pkg/condition"` is the import path.

Dev tip: You can use [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) with your favorite editor, so that it always formats the imports for you.

kyriechen96: Thank you for the explanation! However, I did check the doc and still feel a bit confused, because it only says adding a blank line between each group. For goimports, I tried to run "go install golang.org/x/tools/cmd/goimports@latest" in the terminal of GoLand but nothing happened.

kyriechen96: Should I just click "sort imports"? I saw the option of it.

pkg/resource/parameter_group/update_pre_build_request.go
kyriechen96: Because I wanted to make sure those file names match their hook names(hook call functions from file). Do you think it is better to put all functions into a hooks.go file?

kyriechen96: If merging all functions into a single hooks.go file, the pro is that only one hook file will show up, and the con is that it might be hard to find a specific function which is called by a specific template(hook).

kyriechen96: I found aws-rds team uses only one hooks.go file to include all functions. Currently aws-memorydb controller uses separate files for functions that hooks use. Should I merge them into one hooks.go file?

jaypipes: @kyriechen96 yeah, that's currently the practice that most of the controllers follow...a single `pkg/$resource/hooks.go` file.

kyriechen96: Got it!

pkg/resource/parameter_group/update_pre_build_request.go
kyriechen96: Is a tag pair that only has key but no value acceptable? Currently, I added an If condition statement to run L140 only if value is not empty. Do I need to add an error if input tag pair has no value?

apis/v1alpha1/ack-generate-metadata.yaml
kyriechen96: [](url)> We need to keep the `ack-generate` and `runtime` versions equal (make sure to fetch the code-generator tags `git fetch --tags --all` before the generation)

I've tried the command and regenerated, but the version still shows 5ee0ac0 in the file.

I guess this is the reason. Code-generator shows 5ee0ac0 as version on github now.
Screen Shot 2022-11-02 at 1 23 06 PM

PR #49, opened 2023-01-18
Support Describe Events

kyriechen96:
Issue #, if available:
Resources don't have `Events` field in MemoryDB controller.

Description of changes:
1. Support MemoryDB DescribeEvents for `ACL`, `Cluster`, `User`. Add `Events` field to those resources and add function to sync `Events` from service resource. 
2. Add a custom logic into test framework to validate present of `Events`, and add e2e yaml test file for each resource to validate `Events`.
3. Fix resource name errors in document.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2023-03-03T20:58:18Z"
+  build_date: "2023-03-07T06:28:37Z"
   build_hash: d0f3d78cbea8061f822cbceac3786128f091efe6
   go_version: go1.19
   version: v0.24.2
-api_directory_checksum: 321c79032b9f73aba90926b35167e224f577ecd6
+api_directory_checksum: 863961569e4c45d940e482713a85bac302b37d66
 api_version: v1alpha1
 aws_sdk_go_version: v1.44.93
 generator_config_info:
-  file_checksum: c1d34ab8fe5fc4fa9f0fcc7098afa04a4701e675
+  file_checksum: c6f69226d010aa15f4ed3ac9f2b619bbdcdf798b
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/pkg/resource/acl/sdk.go
+++ b/pkg/resource/acl/sdk.go
@@ -164,6 +164,11 @@ func (rm *resourceManager) sdkFind(
 	}
 
 	rm.setStatusDefaults(ko)
+	ko.Status.Events, err = rm.getEvents(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+
 	if rm.isACLActive(&resource{ko}) {
 		resourceARN := (*string)(ko.Status.ACKResourceMetadata.ARN)
 		tags, err := rm.getTags(ctx, *resourceARN)
--- a/pkg/util/events.go
+++ b/pkg/util/events.go
@@ -1,0 +1,64 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package util
+
+import (
+	"time"
+
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+)
+
+var (
+	// eventsDuration represents the maximum time range worth of
+	//events to retrieve.
+	eventsDuration = time.Duration(14*24) * time.Hour
+	// MaxEvents represents the maximum events can to retrieve.
+	MaxEvents = int64(20)
+)
+
+func NewDescribeEventsInput(
+	sourceName string,
+	sourceType string,
+	maxResults int64,
+) *svcsdk.DescribeEventsInput {
+	input := &svcsdk.DescribeEventsInput{}
+	input.SetSourceType(sourceType)
+	input.SetSourceName(sourceName)
+	input.SetMaxResults(maxResults)
+	input.SetDuration(int64(eventsDuration.Minutes()))
+	return input
+}
+
+// EventsFromDescribe returns a slice of Event structs given the
+// Output response shape from a DescribeEventsWithContext call
+func EventsFromDescribe(
+	resp *svcsdk.DescribeEventsOutput,
+) []*svcapitypes.Event {
+	events := lo.Map(resp.Events, func(respEvent *svcsdk.Event, index int) *svcapitypes.Event {
+		event := &svcapitypes.Event{
+			Message: respEvent.Message,
+		}
+		if respEvent.Date != nil {
+			eventDate := metav1.NewTime(*respEvent.Date)
+			event.Date = &eventDate
+		}
+		return event
+	})
+
+	return events
+}
--- a/test/e2e/scenarios/ACL/acl_create_update.yaml
+++ b/test/e2e/scenarios/ACL/acl_create_update.yaml
@@ -85,8 +85,8 @@ steps:
       UserNames:
         - userone$RANDOM_SUFFIX
         - usertwo$RANDOM_SUFFIX
-  - id: "UPDATE_ACL_WITH_USERNAMES_AND_TAGS"
-    description: "Update userNames and tags"
+  - id: "UPDATE_ACL_USERNAMES_AND_TAGS_AND_VALIDATE_EVENTS"
+    description: "Update userNames and tags, and validate events"
     patch:
       apiVersion: $CRD_GROUP/$CRD_VERSION
       kind: ACL
@@ -105,6 +105,8 @@ steps:
             status: "True"
             timeout: 180
     expect_k8s:
+      status:
+        events:
       spec:
         userNames:
           - userone$RANDOM_SUFFIX
--- a/test/e2e/tests/test_scenarios.py
+++ b/test/e2e/tests/test_scenarios.py
@@ -121,6 +121,14 @@ class UserHelper(helper.ResourceHelper):
     Overrides methods as required for custom resources.
     """
 
+    def assert_extra_items_k8s(self, expected_name, expected_value, k8s_resource) -> bool:
+        # events are processed by custom logic
+        # returns AssertError if events are empty in k8s resource
+        if expected_name == "events":
+            assert k8s_resource.get(expected_name) is not None
+            return True
+        return False
+
     def get_aws_resource(self, resource_name):
         users = self.mdb.describe_users(UserName=resource_name)
         user = users['Users'][0]
@@ -171,6 +179,14 @@ class ACLHelper(helper.ResourceHelper):
     Overrides methods as required for custom resources.
     """
 
+    def assert_extra_items_k8s(self, expected_name, expected_value, k8s_resource) -> bool:
+        # events are processed by custom logic
+        # returns AssertError if events are empty in k8s resource
+        if expected_name == "events":
+            assert k8s_resource.get(expected_name) is not None
+            return True
+        return False
+
     def get_aws_resource(self, resource_name):
         acls = self.mdb.describe_acls(ACLName=resource_name)
         acl = acls['ACLs'][0]
@@ -186,6 +202,14 @@ class ClusterHelper(helper.ResourceHelper):
     Helper for Cluster scenarios.
     Overrides methods as required for custom resources.
     """
+
+    def assert_extra_items_k8s(self, expected_name, expected_value, k8s_resource) -> bool:
+        # events are processed by custom logic
+        # returns AssertError if events are empty in k8s resource
+        if expected_name == "events":
+            assert k8s_resource.get(expected_name) is not None
+            return True
+        return False
 
     def get_aws_resource(self, resource_name):
         clusters = self.mdb.describe_clusters(ClusterName=resource_name)
--- a/apis/v1alpha1/acl.go
+++ b/apis/v1alpha1/acl.go
@@ -54,6 +54,10 @@ type ACLStatus struct {
 	// A list of clusters associated with the ACL.
 	// +kubebuilder:validation:Optional
 	Clusters []*string `json:"clusters,omitempty"`
+	// A list of events. Each element in the list contains detailed information
+	// about one event.
+	// +kubebuilder:validation:Optional
+	Events []*Event `json:"events,omitempty"`
 	// The minimum engine version supported for the ACL
 	// +kubebuilder:validation:Optional
 	MinimumEngineVersion *string `json:"minimumEngineVersion,omitempty"`
--- a/apis/v1alpha1/cluster.go
+++ b/apis/v1alpha1/cluster.go
@@ -124,6 +124,10 @@ type ClusterStatus struct {
 	// The Redis engine patch version used by the cluster
 	// +kubebuilder:validation:Optional
 	EnginePatchVersion *string `json:"enginePatchVersion,omitempty"`
+	// A list of events. Each element in the list contains detailed information
+	// about one event.
+	// +kubebuilder:validation:Optional
+	Events []*Event `json:"events,omitempty"`
 	// The number of shards in the cluster
 	// +kubebuilder:validation:Optional
 	NumberOfShards *int64 `json:"numberOfShards,omitempty"`
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -42,6 +42,11 @@ resources:
         from:
           operation: ListAllowedNodeTypeUpdates
           path: ScaleDownNodeTypes
+      Events:
+        is_read_only: true
+        from:
+          operation: DescribeEvents
+          path: Events
       ParameterGroupName:
         references:
           resource: ParameterGroup
@@ -138,6 +143,11 @@ resources:
         - InvalidParameterValueException
         - InvalidParameterCombinationException
     fields:
+      Events:
+        is_read_only: true
+        from:
+          operation: DescribeEvents
+          path: Events
       UserNames:
         references:
           resource: User
@@ -185,6 +195,11 @@ resources:
         is_secret: true
         compare:
           is_ignored: true
+      Events:
+        is_read_only: true
+        from:
+          operation: DescribeEvents
+          path: Events
     hooks:
       sdk_create_post_set_output:
         code: "rm.setAnnotationsFields(desired, ko)"
--- a/apis/v1alpha1/types.go
+++ b/apis/v1alpha1/types.go
@@ -151,6 +151,7 @@ type Event struct {
 	Date       *metav1.Time `json:"date,omitempty"`
 	Message    *string      `json:"message,omitempty"`
 	SourceName *string      `json:"sourceName,omitempty"`
+	SourceType *string      `json:"sourceType,omitempty"`
 }
 
 // Used to streamline results of a search based on the property being filtered.
--- a/apis/v1alpha1/user.go
+++ b/apis/v1alpha1/user.go
@@ -63,6 +63,10 @@ type UserStatus struct {
 	// Denotes whether the user requires a password to authenticate.
 	// +kubebuilder:validation:Optional
 	Authentication *Authentication `json:"authentication,omitempty"`
+	// A list of events. Each element in the list contains detailed information
+	// about one event.
+	// +kubebuilder:validation:Optional
+	Events []*Event `json:"events,omitempty"`
 	// The minimum engine version supported for the user
 	// +kubebuilder:validation:Optional
 	MinimumEngineVersion *string `json:"minimumEngineVersion,omitempty"`
--- a/apis/v1alpha1/zz_generated.deepcopy.go
+++ b/apis/v1alpha1/zz_generated.deepcopy.go
@@ -204,6 +204,17 @@ func (in *ACLStatus) DeepCopyInto(out *ACLStatus) {
 			}
 		}
 	}
+	if in.Events != nil {
+		in, out := &in.Events, &out.Events
+		*out = make([]*Event, len(*in))
+		for i := range *in {
+			if (*in)[i] != nil {
+				in, out := &(*in)[i], &(*out)[i]
+				*out = new(Event)
+				(*in).DeepCopyInto(*out)
+			}
+		}
+	}
 	if in.MinimumEngineVersion != nil {
 		in, out := &in.MinimumEngineVersion, &out.MinimumEngineVersion
 		*out = new(string)
@@ -805,6 +816,17 @@ func (in *ClusterStatus) DeepCopyInto(out *ClusterStatus) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.Events != nil {
+		in, out := &in.Events, &out.Events
+		*out = make([]*Event, len(*in))
+		for i := range *in {
+			if (*in)[i] != nil {
+				in, out := &(*in)[i], &(*out)[i]
+				*out = new(Event)
+				(*in).DeepCopyInto(*out)
+			}
+		}
+	}
 	if in.NumberOfShards != nil {
 		in, out := &in.NumberOfShards, &out.NumberOfShards
 		*out = new(int64)
@@ -1085,6 +1107,11 @@ func (in *Event) DeepCopyInto(out *Event) {
 	}
 	if in.SourceName != nil {
 		in, out := &in.SourceName, &out.SourceName
+		*out = new(string)
+		**out = **in
+	}
+	if in.SourceType != nil {
+		in, out := &in.SourceType, &out.SourceType
 		*out = new(string)
 		**out = **in
 	}
@@ -2361,6 +2388,17 @@ func (in *UserStatus) DeepCopyInto(out *UserStatus) {
 		in, out := &in.Authentication, &out.Authentication
 		*out = new(Authentication)
 		(*in).DeepCopyInto(*out)
+	}
+	if in.Events != nil {
+		in, out := &in.Events, &out.Events
+		*out = make([]*Event, len(*in))
+		for i := range *in {
+			if (*in)[i] != nil {
+				in, out := &(*in)[i], &(*out)[i]
+				*out = new(Event)
+				(*in).DeepCopyInto(*out)
+			}
+		}
 	}
 	if in.MinimumEngineVersion != nil {
 		in, out := &in.MinimumEngineVersion, &out.MinimumEngineVersion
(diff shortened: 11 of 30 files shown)


helm/templates/deployment.yaml
```diff
@@ -55,6 +55,8 @@ spec:
         - "$(ACK_RESOURCE_TAGS)"
         - --watch-namespace
         - "$(ACK_WATCH_NAMESPACE)"
+        - --deletion-policy
+        - "$(DELETION_POLICY)"
```
RedbackThomson, line 59: You might need to rebase once https://github.com/aws-controllers-k8s/memorydb-controller/pull/48 is merged

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,93 @@
...
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+)
+
+const (
+	// The number of minutes worth of events to retrieve.
+	// 14 days in minutes
+	eventsDuration = 20160
```
RedbackThomson, line 28: Go has a `time` library that is great for time durations. I believe it's:
```
eventsDuration = time.Duration(14) * time.Days
```

Then later you can call
```
eventsDuration.Minutes()
```

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+func (rm *resourceManager) CustomDescribeACLsSetOutput(
```
RedbackThomson, line 31: These functions shouldn't need to be made public. Instead they can be named `customDescribeACLsSetOutput`

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,93 @@
...
+	aclName *string,
+	maxResults int64,
+) ([]*svcapitypes.Event, error) {
+	input := &svcsdk.DescribeEventsInput{}
+	input.SetSourceType("acl")
+	input.SetSourceName(*aclName)
+	input.SetMaxResults(maxResults)
+	input.SetDuration(eventsDuration)
+	resp, err := rm.sdkapi.DescribeEventsWithContext(ctx, input)
+	rm.metrics.RecordAPICall("READ_MANY", "DescribeEvents-ACL", err)
```
RedbackThomson, line 73: For the sake of consistency, I think we should keep the name of the action to just be `DescribeEvents`. We use these traces for the soak testing and I think we want them to map 1:1 with SDK calls

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		rm.log.V(1).Info("Error during DescribeEvents-ACL", "error", err)
+		return nil, err
+	}
+	events := []*svcapitypes.Event{}
+	if resp.Events != nil {
+		for _, respEvent := range resp.Events {
```
RedbackThomson, line 80: I believe `range` still works if `resp.Events` is `nil`, we don't need the nil check.

RedbackThomson (changes requested):
I have some feedback on the ACLs custom code, but it can be applied to each of the other classes as well. Looks like a nice solution though, well done!

pkg/resource/subnet_group/custom_set_output.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+			if respEvent.Date != nil {
+				eventDate := metav1.NewTime(*respEvent.Date)
+				event.Date = &eventDate
+			}
+			events = append(events, event)
+		}
+	}
+	return events, nil
+}
```
jaypipes, line 95: There's a lot of duplication here. Might be good to throw a `pkg/events/events.go` file together with a utility function like this:

```go
// NewDescribeEventsInput returns a populated DescribeEventsInput shape
func NewDescribeEventsInput(
	sourceName *string,
	sourceType string,
	maxResults int64,
) (*svcsdk.DescribeEventsInput {
	input := &svcsdk.DescribeEventsInput{}
	input.SetSourceType(sourceType)
	input.SetSourceName(*sourceName)
	input.SetMaxResults(maxResults)
	input.SetDuration(eventsDuration)
	return input
}

// EventsFromDescribe returns a slice of Event structs given the
// Output response shape from a DescribeEventsWithContext call
func EventsFromDescribe(
	resp *svcsdk.DescribeEventsOutput,
) ([]*svcapitypes.Event) {
	events := []*svcapitypes.Event{}
	if resp.Events != nil {
		for _, respEvent := range resp.Events {
			event := &svcapitypes.Event{}
			if respEvent.Message != nil {
				event.Message = respEvent.Message
			}
			if respEvent.Date != nil {
				eventDate := metav1.NewTime(*respEvent.Date)
				event.Date = &eventDate
			}
			events = append(events, event)
		}
	}
	return events
}
```

and then in all the `pkg/resource/$resource/*.go` files, you could reuse those functions, like so:

```go
import (
...
	"github.com/aws-controllers-k8s/memorydb-controller/pkg/events"
)
...
func (rm *resourceManager) provideEvents(
	ctx context.Context,
	subnetGroupName *string,
	maxResults int64,
) ([]*svcapitypes.Event, error) {
	input := events.NewDescribeEventsInput(subnetGroupName, "subnet-group", maxResults)
	resp, err := rm.sdkapi.DescribeEventsWithContext(ctx, input)
	rm.metrics.RecordAPICall("READ_MANY", "DescribeEvents", err)
	if err != nil {
		rm.log.V(1).Info("Error during DescribeEvents", "error", err)
		return nil, err
	}
	return events.EventsFromDescribe(resp), nil
}
```

That way you minimize the amount of duplicated code, reducing human copy/paste errors.

jaypipes (changes requested):

pkg/resource/subnet_group/custom_set_output.go
RedbackThomson: @jaypipes Is this the first time we've seen custom code placed outside the resource folder?

jaypipes: @RedbackThomson no. Elasticache has had a pkg/common for a while:

https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/pkg/common/delta.go

pkg/resource/acl/custom_set_output.go
kyriechen96: It doesn't have time.Days
Should I use
```
eventsDuration = time.Duration(336) * time.Hour
```
?

kyriechen96: I change to:
```
eventsDuration = time.Duration(336) * time.Hour
```
Then call it and convert to int64 to match the type.
```
int64(eventsDuration.Minutes())
```

pkg/resource/acl/custom_set_output.go
kyriechen96: Corrected.

pkg/resource/acl/custom_set_output.go
kyriechen96: Deleted `if resp.Events != nil`

helm/templates/deployment.yaml
kyriechen96: Did rebase.

pkg/resource/acl/custom_set_output.go
kyriechen96: Yeah changed it to lower case c

pkg/resource/subnet_group/custom_set_output.go
kyriechen96: Created events.go to import and reuse.

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2023-01-10T21:58:28Z"
-  build_hash: 1b20baf45a0b73a11b296050322a384c705fa897
-  go_version: go1.17.5
-  version: v0.22.0
-api_directory_checksum: ee32acc4d4a0ba7e2823dd20fdbe2c4ef1d9e0f4
+  build_date: "2023-01-19T23:23:40Z"
+  build_hash: c6651c200ba136e5c7f50ad8be751fae060a38e6
+  go_version: go1.19
+  version: v0.21.0-9-gc6651c2
```
RedbackThomson: nit: You will need to `git fetch --tags` in your `code-generator` repository, because it doesn't know about the latest version

pkg/events/events.go
```diff
@@ -0,0 +1,49 @@
...
+
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+)
+
+const (
+	// The number of minutes worth of events to retrieve.
+	// 14 days
```
RedbackThomson, line 14: Please update this comment :)

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,67 @@
...
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+	"github.com/aws-controllers-k8s/memorydb-controller/pkg/events"
+)
+
+func (rm *resourceManager) customDescribeACLsSetOutput(
+	ctx context.Context,
+	r *resource,
+	resp *svcsdk.DescribeACLsOutput,
+	ko *svcapitypes.ACL,
+) (*svcapitypes.ACL, error) {
```
RedbackThomson, line 30: Since we are modifying `ko` in these methods, we shouldn't need to return a new `ACL` type. We can just return `error`

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	return ko, nil
+}
+
+func (rm *resourceManager) customSetOutputSupplementAPIs(
+	ctx context.Context,
+	r *resource,
+	ko *svcapitypes.ACL,
+) error {
+	events, err := rm.provideEvents(ctx, r.ko.Spec.Name, 20)
```
RedbackThomson, line 46: Max results is not a required field on the `DescribeEvents` SDK operation, but I agree that we should not have *too many* events in the status or it will become very messy. Could you please make `20` into a constant in the new `pkg/events/events.go` file and add some context around why you chose 20.

pkg/events/events.go
```diff
@@ -0,0 +1,49 @@
...
+	eventsDuration = time.Duration(336) * time.Hour
+)
+
+func NewDescribeEventsInput(
+	sourceName *string,
+	sourceType string,
```
RedbackThomson, line 20: The SDK actually provides an enum for `SourceType`, you can find it in the constants section - https://docs.aws.amazon.com/sdk-for-go/api/service/memorydb/#pkg-constants

So this could be
```suggestion
sourceType svcsdk.SourceType,
```

and then you can pass values to it such as `svcsdk.SourceTypeACL`

RedbackThomson (changes requested):

apis/v1alpha1/ack-generate-metadata.yaml
kyriechen96: Just did it and updated it~

pkg/events/events.go
kyriechen96: Updated ~

pkg/resource/acl/custom_set_output.go
kyriechen96: @RedbackThomson in the operation "Describe", set_output_custom_method generates `ko, err = rm.xxxx` by code generator. If I delete ko in the return value, sdk.go cannot compile.

RedbackThomson: Oh I see. I guess this is fine to leave, for now. It is a bit confusing that we're returning the same object, but it makes more sense than copying it and then returning a modified copy.

jaypipes:
@kyriechen96 It looks like tests are still quite flaky for memorydb. I would suggest we try to address the flakiness before trying `/retest` any more. The test flakiness seems due to indeterminate amount of time for MemoryDB service to get resources into an expected state. Is that correct?

kyriechen96:
@jaypipes Hi our plan now is to finish this feature first, and then fix flaky tests later while doing soak test.

jaypipes:
@kumargauravsharma @nmvk did either of you want to review this PR? I know you want to get this feature in (and the feature looks good to me now after a number of @kyriechen96 changes) however it's still clear to me that the MemoryDB and Elasticache controllers have really flaky tests which makes it very difficult for folks like @kyriechen96 to develop the controller. Have you all considered ways to reduce the test flakiness?

kyriechen96:
@jaypipes I created another PR to fix flaky tests. I just found the root reason for flakiness. I'm going to discuss with them later in the morning.

pkg/events/events.go
kyriechen96: SourceType is not a constant or a field, so I don't find a thing as `svcsdk.SourceType`. This shows red as not defined. However, `svcsdk.SourceTypeAcl`, `SourceTypeCluster`, `SourceTypeUser` work.
https://pkg.go.dev/github.com/aws/aws-sdk-go/service/memorydb@v1.44.93
https://github.com/aws/aws-sdk-go/blob/v1.44.93/service/memorydb/api.go#L12248

pkg/util/events.go
```diff
@@ -0,0 +1,51 @@
...
+	// eventsDuration represents the maximum time range worth of
+	//events to retrieve.
+	eventsDuration = time.Duration(336) * time.Hour
+	// MaxEvents represents the maximum events can to retrieve.
+	MaxEvents = int64(20)
+)
+
+func NewDescribeEventsInput(
+	sourceName *string,
+	sourceType string,
```
A-Hilaly, line 22: Why take one string and one string pointer

test/e2e/scenarios/ACL/acl_create_update.yaml
```diff
@@ -102,6 +102,8 @@ steps:
             status: "True"
             timeout: 180
     expect_k8s:
+      status:
+        events:
```
A-Hilaly, line 106: I don't see any events in the scenarios. Is it intentional?

pkg/util/events.go
```diff
@@ -0,0 +1,51 @@
...
+// Output response shape from a DescribeEventsWithContext call
+func EventsFromDescribe(
+	resp *svcsdk.DescribeEventsOutput,
+) []*svcapitypes.Event {
+	var events []*svcapitypes.Event
+	for _, respEvent := range resp.Events {
+		event := &svcapitypes.Event{}
+		if respEvent.Message != nil {
+			event.Message = respEvent.Message
+		}
```
A-Hilaly, line 43: nit:
```suggestion
        event.Message = respEvent.Message
```

pkg/util/events.go
```diff
@@ -0,0 +1,51 @@
+package util
```
A-Hilaly, line 1: You need the license boilerplate at the beginning of new Go files https://github.com/aws-controllers-k8s/memorydb-controller/blob/main/cmd/controller/main.go#L1-L12

pkg/resource/acl/sdk.go
```diff
@@ -164,6 +164,11 @@ func (rm *resourceManager) sdkFind(
 	}
 
 	rm.setStatusDefaults(ko)
+	// custom set output from response
+	ko, err = rm.customDescribeACLsSetOutput(ctx, r, resp, ko)
+	if err != nil {
+		return nil, err
+	}
```
A-Hilaly, line 171: I'm not super fan of this function's signature. Can you change it to take the minimum possible of arguments.
You could also return the events instead of the `ko` and do something like
```
ko.Status.Events, err = rm.customDescribeACLsSetOutput(ctx, r, resp)
```

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,67 @@
...
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+	svcutil "github.com/aws-controllers-k8s/memorydb-controller/pkg/util"
+)
+
+func (rm *resourceManager) customDescribeACLsSetOutput(
+	ctx context.Context,
+	r *resource,
+	resp *svcsdk.DescribeACLsOutput,
+	ko *svcapitypes.ACL,
+) (*svcapitypes.ACL, error) {
```
A-Hilaly, line 30: Just get rid of the `r` and `ko` in the same function arguemnts. You can just using `r` to pick up the name and return the events.

pkg/resource/acl/custom_set_output.go
```diff
@@ -0,0 +1,67 @@
...
+	if len(resp.ACLs) == 0 {
+		return ko, nil
+	}
```
A-Hilaly, line 33: Since `resp` is only used here. I think this check should go to the describe hook. which can reduce the number of parameters needed for this function:
```
func (rm *resourceManager) customDescribeACLsSetOutput(
	ctx context.Context,
	r *resource,
) ([]*events..., error) {
```

pkg/resource/acl/sdk.go
A-Hilaly: Also at this stage you know that, `len(resp.ACL) == 1` so no need to check the length at all inside `customDescribeACLsSetOutput`

A-Hilaly (changes requested):
Good start. Left few comments/questions

pkg/util/events.go
kyriechen96: They match the parameters in `provideEvents` function in which it calls this `NewDescribeEventsInput` function.

kyriechen96: Should I keep all parameters in `NewDescribeEventsInput` as string?

test/e2e/scenarios/ACL/acl_create_update.yaml
kyriechen96: Because `events` are generated by service according to behaviors of a resource. We cannot predict the content of `events`. Hence, the custom logic adding to test framework only validate if there is something inside of `events` instead of validating what exactly in `events`.

kyriechen96: Also, because of the way `runtime` works, k8s resource always has one less event than aws resource.

pkg/util/events.go
kyriechen96: Added.

pkg/resource/acl/sdk.go
kyriechen96: Yeah, I forgot that `resp` is checked before in `sdkFind`

kyriechen96: `ko, err = rm.customDescribeACLsSetOutput(ctx, r, resp, ko)` this is generated by `code-generator`. I didn't write it in `templates`, so I cannot change it.

kyriechen96: If you want to do that way, I can move them to `sdk_read_many_post_set_output.go.tpl` instead of using `set_output_custom`

pkg/resource/acl/custom_set_output.go
kyriechen96: This is from required parameter format by `code-generator`, but I can do the way you want as custom code in `templates`.

pkg/resource/acl/custom_set_output.go
kyriechen96: Same as above.

pkg/resource/acl/sdk.go
A-Hilaly: Ohhh i didn't know that a generated signature. Yes i think `sdk_read_many_post_set_output.go.tpl` would be better here

pkg/util/events.go
kyriechen96: Fixed.

pkg/util/events.go
kyriechen96: Modified.

pkg/resource/acl/custom_set_output.go
kyriechen96: Moved to `sdk_read_many_post_set_output.go.tpl`.

pkg/resource/acl/custom_set_output.go
kyriechen96: Moved to `sdk_read_many_post_set_output.go.tpl` and modified.

test/e2e/tests/test_scenarios.py
```diff
@@ -121,6 +121,14 @@ class UserHelper(helper.ResourceHelper):
     Overrides methods as required for custom resources.
     """
 
+    def assert_extra_items_k8s(self, expected_name, expected_value, k8s_resource) -> bool:
+        # events are processed by custom logic
+        # returns AssertError if events are empty in k8s resource
+        if expected_name == "events":
+            assert k8s_resource.get(expected_name) is not None
+            return True
```
RedbackThomson, line 129: This condition can be added to `ResourceHelper` so it's not duplicated between all of the classes?

test/e2e/scenarios/ACL/acl_create_update.yaml
RedbackThomson: If we add `requeue_on_success_seconds`, then we can continue to populate events as they become available. See how SageMaker uses it - https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/generator.yaml#L64-L65

The docs have a section about this, so you can understand what it does - https://aws-controllers-k8s.github.io/community/docs/contributor-docs/code-generator-config/#reconcile-controlling-reconciliation-and-requeue-logic

pkg/util/events.go
```diff
@@ -0,0 +1,63 @@
...
+		event := &svcapitypes.Event{}
+		event.Message = respEvent.Message
+
+		if respEvent.Date != nil {
+			eventDate := metav1.NewTime(*respEvent.Date)
+			event.Date = &eventDate
+		}
+		events = append(events, event)
+	}
+	return events
```
RedbackThomson, line 62: nit: This mapping of `resp.Event` to `svcapitypes.Event` can be made even simpler using the `samber/lo` package's `lo.Map`. Try something like this:
```suggestion
	events := lo.Map(resp.Events, func(respEvent *svcsdk.Event, index int) *svcapitypes.Event {
		event := &svcapitypes.Event{
			Message: respEvent.Message
		}

		if respEvent.Date != nil {
			eventDate := metav1.NewTime(*respEvent.Date)
			event.Date = &eventDate
		}

		return event
	})
```

test/e2e/tests/test_scenarios.py
kyriechen96: It is only for `cluster`, `acl`, and `user` resources. MemoryDB service doesn't generate `events` for other three resources.

RedbackThomson: Oh. Sure.

pkg/util/events.go
kyriechen96: Modified.

pkg/util/events.go
jljaco: Sorry, what's the rationale here for having a string and a string pointer?  It seems much cleaner to adopt one convention for both parameters and to adjust the places this method is called accordingly.

pkg/util/events.go
```diff
@@ -0,0 +1,63 @@
...
+	input := &svcsdk.DescribeEventsInput{}
+	input.SetSourceType(sourceType)
+	input.SetSourceName(*sourceName)
+	input.SetMaxResults(maxResults)
+	input.SetDuration(int64(eventsDuration.Minutes()))
+	return input
+}
+
+// EventsFromDescribe returns a slice of Event structs given the
+// Output response shape from a DescribeEventsWithContext call
```
jljaco, line 47: Can you please point me to the documentation for `DescribeEventsWithContext`?

pkg/util/events.go
```diff
@@ -0,0 +1,63 @@
...
+func EventsFromDescribe(
+	resp *svcsdk.DescribeEventsOutput,
+) []*svcapitypes.Event {
+	var events []*svcapitypes.Event
+	for _, respEvent := range resp.Events {
```
jljaco, line 52: Why are you omitting `SourceName` and `SourceType` from this copy operation?

test/e2e/scenarios/ACL/acl_create_update.yaml
kyriechen96: Added.

pkg/util/events.go
kyriechen96: I checked online document before and `Event` there didn't have `SourceName` and `SourceType` when I wrote this code. I see `Event` has them in `types.go` file now, and I will add them!
https://pkg.go.dev/github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1#Event

kyriechen96: Added~

RedbackThomson: `SourceName` will always be the name of the `User`/`ACL`/`Snapshot` and `SourceType` will be `"User"`,`"ACL",`"Snapshot"`? This information is already assumed based on the resource that has the events. Do we need to repeat this in every event?

kyriechen96: @RedbackThomson I used debugger and debugger showed it doesn't auto fill `SourceType` and `SourceName`

test/e2e/scenarios/ACL/acl_create_update.yaml
kyriechen96: Removed `requeue_on_success_seconds`. It caused a lot of panic and made a lot of tests failed.

pkg/util/events.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+)
+
+var (
+	// eventsDuration represents the maximum time range worth of
+	//events to retrieve.
+	eventsDuration = time.Duration(336) * time.Hour
```
A-Hilaly, line 29: nit:
```suggestion
    eventsDuration = time.Duration(14*24) * time.Hour
```

A-Hilaly (approved):
Nice and neat! Good work on this @kyriechen96

pkg/util/events.go
kyriechen96: Modified~

pkg/util/events.go
RedbackThomson: That's not what I meant. I meant that the values returned by `DescribeEvents` will always have the same `SourceType`. If we called `DescribeEvents` for `ACL` all `SourceType` in the response will be `"ACL"`. When we look at the status of `ACL` it's not useful to have `SourceType`, since we already know the event is for the `ACL` object that we are describing

kyriechen96: I got you. For me, I think it doesn't matter to have `SourceType` or `SourceName` for `events` in a resource because we can get that information from resource `spec`. However, since we do have these two fields in `events`, so I guess it is better to fill them and not leave them to empty?

kyriechen96: The reason `SourceType` and `SourceName` exist is that MemoryDB API calls `describe-events` that needs to know which specific resource customer is calling to check events. However, in k8s, customer just calls `kubectl describe resource` and the `events` is a field in the `status` of resource, so `SourceType` and `SourceName` don't impact in this case.

RedbackThomson: Right. If we don't set them in this method, does `sourceType` and `sourceName` show up as empty strings in the status or do they just not appear at all? If they show up as empty strings then I think we should include them, or it will look weird. But if they don't appear, I think it's better to leave them out so that it doesn't clutter the status with useless information

kyriechen96: Let me test it now

kyriechen96: If I leave `sourceType` and `sourceName` to empty, they won't show up. Decide to keep them empty.

test/e2e/scenarios/ACL/acl_create_update.yaml
RedbackThomson: Oh no. Okay we can revisit this some other time

RedbackThomson (approved):
/lgtm

PR #51, opened 2023-01-23
Modify tags related functions and some test files to fix flaky assertion errors

kyriechen96:
Issue #, if available:
MemoryDB Controller has some flaky tests that randomly show assertion error.
Root causes:
1. Snapshot test files didn't clean resources which were created from previous failed tests, so eventually test account reached node limit.
2. Delta compare considers order matters, so it considers two tag arrays are different even though they have same elements with different orders. Function of updating tags didn't include this corner case, and it caused infinite update.
3. When creating resources, some resources take a while to finish creation. Function of getting tags got error because it couldn't find resource, and this error covered previous ResourceNotFound error. Hence, reconciler didn't call requeue, because it didn't recognize the new error.
4. Cluster update validation test file has a step to update and validate Snapshot Window (daily time range to auto snapshot). If this step is executing within the time range of new Snapshot Window, update would succeed (ResourceSynced returns true) because two resources match, but cluster starts snapshotting after it's Snapshot Window is updated to the current time. Hence, for next step to update other fields of cluster, new fields couldn't be updated on time, because the status of cluster is snapshotting and cluster cannot be updated.

Description of changes:
1. Fix updateTags function and it covers situation of equal tag arrays. (It still wastes time to call sdkUpdate over and over. It is better to modify delta comparison of tags in code-generator. New PR is created (https://github.com/aws-controllers-k8s/community/issues/1658).
2. Change updateTags functions from parameter group and subnet group to match the updateTags functions in other resources.
3. Add a status check for getTags. It is only called when status of resource (cluster, snapshot, acl, user) is active.
4. Increase wait time for the next update step after Snapshot Window update. The new wait time is enough to cover snapshotting process.
5. Use same method to create cluster for snapshot for snapshot_validate_tags.yaml test file.
6. Correct user names and delete unused resources in test files.
7. Modify some test IDs, step IDs, and descriptions which are not appropriate that makes debugging really hard.
8. Fix import order of hook files.
9. Modify equalStrings function.


test/e2e/scenarios/Cluster/cluster_terminal_condition.yaml
```diff
@@ -32,6 +32,9 @@ steps:
         name: cluster$RANDOM_SUFFIX
       spec:
         nodeType: db.t4g.small
+        aclName: open-access
+        numShards: 1
+        numReplicasPerShard: 0
```
nmvk, line 37: This should not be added whole idea is test should run once you fix the spec that has issue?

test/e2e/scenarios/Cluster/cluster_terminal_condition.yaml
```diff
@@ -13,9 +13,9 @@ steps:
         name: cluster$RANDOM_SUFFIX
       spec:
         name: cluster$RANDOM_SUFFIX
-        nodeType: cache.t4g.medium
+        nodeType: cache.t4g.small
         aclName: open-access
-        numShards: 2
+        numShards: 1
```
nmvk, line 18: I dont think these changes are needed. `cache` is not valid node type for memorydb

test/e2e/scenarios/ACL/acl_create_update.yaml
```diff
@@ -58,12 +58,12 @@ steps:
         userNames:
           - userone$RANDOM_SUFFIX
           - usertwo$RANDOM_SUFFIX
-    wait:
```
nmvk, line 61:
```
wait:
      status:
        conditions:
          ACK.ResourceSynced:
            status: "True"
            timeout: 180
```

Why would hard coding the value give advantage over status change wait? I prefer the previous approach as it helps us catch any issue with status change flipping back and forth.

nmvk (changes requested):

test/e2e/scenarios/ACL/acl_create_update.yaml
kyriechen96: Previous approach causes random assertion error. My assumption is there is something wrong with previous approach, and I modified to the current approach to check whether it works. And I think we need to investigate the root cause.

test/e2e/scenarios/Cluster/cluster_terminal_condition.yaml
kyriechen96: Ideally I agree with you, but this causes node limit exceed. It doesn't have this issue before, but this happens recently. It may cause by some code changes on service side. The modification makes test pass now, so we need to investigate the root cause.

test/e2e/scenarios/Cluster/cluster_terminal_condition.yaml
kyriechen96: Yeah the root reason is that test account is full now. Deleting some clusters in account.

jaypipes:
@kyriechen96 Please see the error from the tests:

```
AssertionError: Condition status mismatch. Expected condition: ACK.ResourceSynced - {'status': 'True'} but found {'lastTransitionTime': '2023-01-26T17:22:26Z', 'message': 'Unable to determine if desired resource state matches latest observed state', 'reason': 'ACLNotFoundFault: acl-jl1ptly9i46hs93 is either not present or not available.', 'status': 'Unknown', 'type': 'ACK.ResourceSynced'}
```

The issue isn't timeouts. The issue seems to be a missing ACL dependency in one of the tests I think?

kyriechen96:
> @kyriechen96 Please see the error from the tests:
> 
> ```
> AssertionError: Condition status mismatch. Expected condition: ACK.ResourceSynced - {'status': 'True'} but found {'lastTransitionTime': '2023-01-26T17:22:26Z', 'message': 'Unable to determine if desired resource state matches latest observed state', 'reason': 'ACLNotFoundFault: acl-jl1ptly9i46hs93 is either not present or not available.', 'status': 'Unknown', 'type': 'ACK.ResourceSynced'}
> ```
> 
> The issue isn't timeouts. The issue seems to be a missing ACL dependency in one of the tests I think?

We had a discussion of it. This is happened during condition check for ACL creation. I'm fixing it now.

test/e2e/scenarios/Cluster/cluster_create_update.yaml
```diff
@@ -61,17 +61,16 @@ steps:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 7200
-
-  - id: "USER_ONE_CREATE"
-    description: "Create User ONE"
+            timeout: 3600
```
RedbackThomson, line 64: 60 minutes? Is this an appropriate value?

test/e2e/scenarios/Cluster/cluster_create_update.yaml
```diff
@@ -137,8 +134,7 @@ steps:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 100
-
+            timeout: 180
```
RedbackThomson, line 137: 3 minutes to create a parameter group? My basic tests show that they are created practically instantly when using the command line. Do you know where the extra time is spent before it reaches `ResourceSynced`?

test/e2e/scenarios/Cluster/cluster_create_update.yaml
```diff
@@ -153,16 +149,14 @@ steps:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 7200
-
+            timeout: 180
```
RedbackThomson, line 152: 180 is a lot better than 7200, but this is essentially an instant operation too. I'd expect a much smaller timeout. Like 10 seconds.

test/e2e/scenarios/Snapshot/snapshot_create_terminal_condition.yaml
```diff
@@ -13,11 +13,12 @@ steps:
...
         name: snapshot$RANDOM_SUFFIX
-    wait: 600
-    expect:
+    wait:
       status:
         conditions:
-          ACK.Terminal: "True"
+          ACK.Terminal:
+            status: "True"
+            timeout: 600
```
RedbackThomson, line 21: It takes like 2 seconds to call `CreateSnapshot` and the response is practically immediate. It shouldn't take 10 minutes to reach the condition, more like 5 seconds.

test/e2e/scenarios/Snapshot/snapshot_update_with_tags.yaml
```diff
@@ -56,7 +38,7 @@ steps:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 100
+            timeout: 180
```
RedbackThomson, line 41: This can be 5 seconds and it should still work. Every tag operation can be 5 seconds IMO

RedbackThomson:
I'm not super familiar with the memorydb snapshotting process. Would you mind explaining this a little more:
> If this step is executing within the time range of new Snapshot Window, update would succeed (ResourceSynced returns true) because two resources match, but cluster starts snapshotting after it's Snapshot Window is updated to the current time. Hence, for next step to update other fields of cluster, new fields couldn't be updated on time, because the status of cluster is snapshotting and cluster cannot be updated

pkg/resource/acl/hooks.go
```diff
@@ -41,6 +41,14 @@ func (rm *resourceManager) validateACLNeedsUpdate(
 	return nil
 }
 
+// aclActive returns true when the status of the given ACL is set to `active`
+func (rm *resourceManager) aclActive(
+	latest *resource,
+) bool {
+	latestStatus := latest.ko.Status.Status
+	return latestStatus != nil && *latestStatus == "active"
```
RedbackThomson, line 49: I see there is no constant in the SDK for the `active` status - would you mind moving this to a variable at the top of the file so it's not a magic string.

pkg/resource/acl/sdk.go
```diff
@@ -164,12 +164,14 @@ func (rm *resourceManager) sdkFind(
 	}
 
 	rm.setStatusDefaults(ko)
-	resourceARN := (*string)(ko.Status.ACKResourceMetadata.ARN)
-	tags, err := rm.getTags(ctx, *resourceARN)
-	if err != nil {
-		return nil, err
+	if rm.aclActive(&resource{ko}) {
```
RedbackThomson, line 167: If not active, lets set it to `resourceSynced = false` condition so that it continues to requeue until it is active.

pkg/resource/cluster/hooks.go
```diff
@@ -264,6 +264,14 @@ func (rm *resourceManager) newMemoryDBClusterUploadPayload(
 	return res
 }
 
+// clusterActive returns true when the status of the given Cluster is set to `active`
+func (rm *resourceManager) clusterActive(
+	latest *resource,
+) bool {
+	latestStatus := latest.ko.Status.Status
+	return latestStatus != nil && *latestStatus == "active"
```
RedbackThomson, line 272: Same as ACL, into variable please

pkg/resource/acl/hooks.go
```diff
@@ -114,17 +122,23 @@ func computeTagsDelta(
 	latest []*svcapitypes.Tag,
```
RedbackThomson, line 122: It looks like we are reusing the `computeTagsDelta` in every `hooks.go` file. We can reduce this duplication by moving this method into a package that is accessible by all resources. RDS puts theirs in `pkg/utils/tags.go` and then imports that module inside `hooks.go` - https://github.com/aws-controllers-k8s/rds-controller/blob/main/pkg/util/tags.go

test/e2e/replacement_values.py
```diff
@@ -24,5 +24,6 @@
     "TOPIC2": get_bootstrap_resources().Topic2.topic_arn,
     "KMSKEY": get_bootstrap_resources().KMSKey.key,
     "SNAPSHOT_CLUSTER_NAME1": get_bootstrap_resources().Cluster1.clusterName,
-    "SNAPSHOT_CLUSTER_NAME2": get_bootstrap_resources().Cluster2.clusterName
+    "SNAPSHOT_CLUSTER_NAME2": get_bootstrap_resources().Cluster2.clusterName,
+    "SNAPSHOT_CLUSTER_NAME3": get_bootstrap_resources().Cluster3.clusterName
```
RedbackThomson, line 28: How long does it take to stand up 3 clusters? Is there any way to reduce this down to 1 - or at most 2? This seems really expensive and slow

kyriechen96:
> I'm not super familiar with the memorydb snapshotting process. Would you mind explaining this a little more:
> 
> > If this step is executing within the time range of new Snapshot Window, update would succeed (ResourceSynced returns true) because two resources match, but cluster starts snapshotting after it's Snapshot Window is updated to the current time. Hence, for next step to update other fields of cluster, new fields couldn't be updated on time, because the status of cluster is snapshotting and cluster cannot be updated

There is a daily snapshotting behavior for clusters. Cluster start daily snapshotting when the time arrives Snapshot Window. 

For example, if Snapshot Window of a cluster is 8:00am - 9:00am, cluster starts to snapshot at 8:00am everyday. 

In the case of test, we update Snapshot Window to 8:00am - 9:00am. If this step runs during 8am - 9am, firstly it tells update is successful (return true for ResourceSynced) and then it starts snapshotting because the current time is in between of 8am and 9am.

RedbackThomson (changes requested):
Amazing work untangling all of those missing pieces. I'm glad solving flaky tests wasn't just a matter of increasing timeouts - you actually got to the root cause of quite a few nagging bugs. I've left quite a few minor comments in the PR. 

As a follow up, I'd like it if we could have another PR where we reduce the timeouts for operations that are waiting for a needlessly long time. `TagResource` and `UntagResource` are instant operations, so updating an objects tags shouldn't wait for 30 seconds. But this PR is already getting long, so we can do that next time.

test/e2e/scenarios/Cluster/cluster_create_update.yaml
kyriechen96: It has a rare chance that previous update of snapshot window causes a snapshotting process which could take a long time. All updates have to wait till cluster finishes snapshotting.

test/e2e/scenarios/Cluster/cluster_create_update.yaml
kyriechen96: Will create a new PR to fix all timeout.

test/e2e/scenarios/Cluster/cluster_create_update.yaml
kyriechen96: Will create a new PR to fix all timeout.

test/e2e/scenarios/Snapshot/snapshot_create_terminal_condition.yaml
kyriechen96: Will create a new PR to fix all timeout.

test/e2e/scenarios/Snapshot/snapshot_update_with_tags.yaml
kyriechen96: Will create a new PR to fix all timeout.

pkg/resource/acl/sdk.go
kyriechen96: We cannot, it causes a bug.

test/e2e/replacement_values.py
kyriechen96: I don't think we can. Those clusters are made for validating snapshots. All tests run at the mean time. If they use same cluster, it has a problem that one test finishes and want to delete cluster for cleaning but another test is still using this cluster.

pkg/resource/acl/hooks.go
kyriechen96: Yeah, we can modify it after using latest code-generator to generate again.

test/e2e/scenarios/Cluster/cluster_create_update.yaml
kyriechen96: Set it back to 100 to keep it same, and will have a new PR for finding a more appropriate value.

test/e2e/scenarios/Snapshot/snapshot_update_with_tags.yaml
kyriechen96: Set it back to 100. Will find a more appropriate value in a new PR.

pkg/resource/acl/hooks.go
kyriechen96: Updated

pkg/resource/cluster/hooks.go
kyriechen96: Updated.

pkg/resource/acl/hooks.go
kyriechen96: I will create another PR to do it.

pkg/resource/acl/sdk.go
kyriechen96: It causes failure for all snapshot and cluster creations.

RedbackThomson: Nevermind, you already have the logic for using the status to determine `IsSynced` in your `generator.yaml`. No need to write custom code when we already support that. Great

test/e2e/scenarios/Snapshot/snapshot_update_with_tags.yaml
```diff
@@ -4,33 +4,15 @@ description: "In this test we create snapshot and update tags"
...
+  - id: "CREATE_INITIAL_SNAPSHOT"
+    description: "Create Initial Snapshot"
     create:
       apiVersion: $CRD_GROUP/$CRD_VERSION
       kind: Snapshot
       metadata:
         name: snapshot$RANDOM_SUFFIX
       spec:
-        clusterName: cluster$RANDOM_SUFFIX
+        clusterName: $SNAPSHOT_CLUSTER_NAME3
```
nmvk, line 15: I do not think we need to do this in bootstrap resource this should not cause flakiness.

test/e2e/replacement_values.py
nmvk: I agree with @RedbackThomson, it is expensive to do here we can combine tagging validation as part of other snapshotting tests. For example one cluster can be snapshotted any number of times.

nmvk (changes requested):

test/e2e/scenarios/Snapshot/snapshot_update_with_tags.yaml
kyriechen96: @nmvk I can change it back, but may I know why we don't use Cluster(Bootstrappable) if it is how we create cluster for Ack snapshot resource testing?

test/e2e/replacement_values.py
kyriechen96: @nmvk tagging validation test is for Tag and Untag APIs. I'm not sure whether we should merge it into other snapshot tests. Controller has tagging validation test for each resource. I don't think it is a good idea to only merge tags validation test for snapshot one into other snapshotting tests. If you really think it is better to reduce cluster numbers, then I will create a new PR for it.

test/e2e/scenarios/Snapshot/snapshot_update_with_tags.yaml
kyriechen96: Either use bootstrap resource or create a new cluster does the same thing. I thought it is better to use bootstrap resource because all other snapshot tests use this way.

kyriechen96: Btw, I will change it back and create a new PR for it (or merge it into other snapshot tests).

test/e2e/scenarios/Snapshot/snapshot_create.yaml
```diff
@@ -27,3 +27,10 @@ steps:
       kind: Snapshot
       metadata:
         name: snapshot$RANDOM_SUFFIX
+  - id: "DELETE_CLUSTER"
+    description: "Delete cluster"
+    delete:
+      apiVersion: $CRD_GROUP/$CRD_VERSION
+      kind: Cluster
+      metadata:
+        name: $SNAPSHOT_CLUSTER_NAME1
```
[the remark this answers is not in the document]
kyriechen96: You mean in `service_cleanup() `? I didn't realize we have deletion there. I will remove this part.

pkg/resource/cluster/hooks.go
```diff
@@ -337,17 +346,23 @@ func computeTagsDelta(
...
 				if !equalStrings(latestElement.Value, desiredElement.Value) {
 					addedOrUpdated = append(addedOrUpdated, desiredElement)
 				}
-				continue
+				break
 			}
 		}
+		if hasSameKey {
+			continue
+		}
```
A-Hilaly, line 365: I see that you wanna ignore duplicated keys +++

pkg/resource/cluster/hooks.go
```diff
@@ -30,8 +30,9 @@ import (
 )
 
 var (
-	condMsgCurrentlyDeleting     = "cluster currently being deleted"
-	condMsgNoDeleteWhileUpdating = "cluster is being updated. cannot delete"
+	condMsgCurrentlyDeleting            = "cluster currently being deleted"
+	condMsgNoDeleteWhileUpdating        = "cluster is being updated. cannot delete"
```
A-Hilaly, line 34: Can we make messages a bit consistent? `cluster currently being updated`

pkg/resource/parameter_group/hooks.go
```diff
@@ -15,6 +15,7 @@ package parameter_group
 
 import (
 	"context"
+	ackutil "github.com/aws-controllers-k8s/runtime/pkg/util"
```
A-Hilaly, line 18: https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/resource/parameter_group/hooks.go
```diff
@@ -277,3 +271,23 @@ func sdkTagsFromResourceTags(
...
+	}
+	return tags
+}
+
+func equalStrings(a, b *string) bool {
+	if a == nil {
+		return b == nil || *b == ""
+	}
+	return (*a == "" && b == nil) || *a == *b
+}
```
A-Hilaly, line 293: This might not panic in the tags case, it could with other fields. Correct implementation here: https://github.com/aws-controllers-k8s/community/issues/1654

pkg/resource/snapshot/hooks.go
```diff
@@ -17,16 +17,20 @@ import (
...
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	ackutil "github.com/aws-controllers-k8s/runtime/pkg/util"
+	svcsdk "github.com/aws/aws-sdk-go/service/memorydb"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
+)
```
A-Hilaly, line 30: :+1:

pkg/resource/subnet_group/hooks.go
```diff
@@ -15,6 +15,7 @@ package subnet_group
 
 import (
 	"context"
+	ackutil "github.com/aws-controllers-k8s/runtime/pkg/util"
```
A-Hilaly, line 18: ditto

A-Hilaly (commented):
Great stuff Kyrie!

pkg/resource/cluster/hooks.go
kyriechen96: These two variables are not related to flaky tests. I didn't modify or create them in this PR, so I will mark it and change it in a new PR.

pkg/resource/parameter_group/hooks.go
kyriechen96: Yeah it was automatically added there. I will change the order manually.

pkg/resource/parameter_group/hooks.go
kyriechen96: I will fix it for all resource hooks in the new PR.

pkg/resource/parameter_group/hooks.go
kyriechen96: Actually I will fix it in the next PR. Looks like most hooks have import order issue.

pkg/resource/subnet_group/hooks.go
kyriechen96: I'll fix it in next PR.

pkg/resource/parameter_group/hooks.go
RedbackThomson: You should be able to just copy the code from that issue into all of your `equalStrings` functions. It would be better if we address it now than leave it for another PR, where we might forget.

pkg/resource/subnet_group/hooks.go
kyriechen96: Fixed import orders for all hooks.

pkg/resource/parameter_group/hooks.go
kyriechen96: Fixed it for all hook files.

pkg/resource/cluster/hooks.go
RedbackThomson: Yes please. The english grammar for these two messages should be improved. `cluster is currently being deleted` and `cluster is currently being updated, cannot delete` are my suggestions

pkg/resource/user/hooks.go
```diff
@@ -55,6 +59,14 @@ func (rm *resourceManager) validateUserNeedsUpdate(
 	return nil, nil
 }
 
+// userActive returns true when the status of the given User is set to `active`
+func (rm *resourceManager) userActive(
```
RedbackThomson, line 63: nit: methods returning booleans should start with an article. For example `isUserActive`

RedbackThomson (approved):
I think this is ready to go now! Running the soak tests will be the real determination of whether all of these tests are solid, but for now this looks good.

I've left one nit that's applicable to all of the `<resource>Active` hook methods.

nmvk (approved):

pkg/resource/user/hooks.go
A-Hilaly: @kyriechen96 wants to address this in a seperate PR

PR #56, opened 2023-02-07
Fix incorrect status of cluster and snapshot

kyriechen96:
Issue #, if available:
1. `status` of cluster and snapshot should be `available` instead of `active`.
2. names of function for checking `status` of resources are not proper.

Description of changes:
1. Add `synced` of snapshot into generator. Correct `status` of cluster and snapshot from `active` to `available`.
2. Uncomment some parts in two yaml test files related to tags update of cluster and snapshot.
3. Modify functions for checking `status` of cluster, snapshot, acl, user to proper names.

### DIFF
--- a/pkg/resource/acl/hooks.go
+++ b/pkg/resource/acl/hooks.go
@@ -45,8 +45,8 @@ func (rm *resourceManager) validateACLNeedsUpdate(
 	return nil
 }
 
-// aclActive returns true when the status of the given ACL is set to `active`
-func (rm *resourceManager) aclActive(
+// isACLActive returns true when the status of the given ACL is set to `active`
+func (rm *resourceManager) isACLActive(
 	latest *resource,
 ) bool {
 	latestStatus := latest.ko.Status.Status
--- a/pkg/resource/cluster/hooks.go
+++ b/pkg/resource/cluster/hooks.go
@@ -31,7 +31,7 @@ import (
 var (
 	condMsgCurrentlyDeleting            = "cluster currently being deleted"
 	condMsgNoDeleteWhileUpdating        = "cluster is being updated. cannot delete"
-	resourceStatusActive         string = "active"
+	resourceStatusAvailable      string = "available"
 )
 
 var (
@@ -264,12 +264,12 @@ func (rm *resourceManager) newMemoryDBClusterUploadPayload(
 	return res
 }
 
-// clusterActive returns true when the status of the given Cluster is set to `active`
-func (rm *resourceManager) clusterActive(
+// isClusterAvailable returns true when the status of the given Cluster is set to `available`
+func (rm *resourceManager) isClusterAvailable(
 	latest *resource,
 ) bool {
 	latestStatus := latest.ko.Status.Status
-	return latestStatus != nil && *latestStatus == resourceStatusActive
+	return latestStatus != nil && *latestStatus == resourceStatusAvailable
 }
 
 // getTags gets tags from given ParameterGroup.
--- a/pkg/resource/snapshot/hooks.go
+++ b/pkg/resource/snapshot/hooks.go
@@ -29,7 +29,7 @@ import (
 )
 
 var (
-	resourceStatusActive string = "active"
+	resourceStatusAvailable string = "available"
 )
 
 func (rm *resourceManager) customDescribeSnapshotSetOutput(
@@ -251,12 +251,12 @@ func (rm *resourceManager) newCopySnapshotPayload(
 	return res, nil
 }
 
-// snapshotActive returns true when the status of the given Snapshot is set to `active`
-func (rm *resourceManager) snapshotActive(
+// isSnapshotAvailable returns true when the status of the given Snapshot is set to `available`
+func (rm *resourceManager) isSnapshotAvailable(
 	latest *resource,
 ) bool {
 	latestStatus := latest.ko.Status.Status
-	return latestStatus != nil && *latestStatus == resourceStatusActive
+	return latestStatus != nil && *latestStatus == resourceStatusAvailable
 }
 
 // getTags gets tags from given ParameterGroup.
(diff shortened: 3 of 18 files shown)


pkg/resource/acl/hooks.go
```diff
@@ -45,8 +45,8 @@ func (rm *resourceManager) validateACLNeedsUpdate(
 	return nil
 }
 
-// aclActive returns true when the status of the given ACL is set to `active`
-func (rm *resourceManager) aclActive(
+// isAclActive returns true when the status of the given ACL is set to `active`
+func (rm *resourceManager) isAclActive(
```
RedbackThomson, line 49: nit: use the proper casing
```suggestion
func (rm *resourceManager) isACLActive(
```

RedbackThomson (commented):
Just one nit but otherwise a good catch. Do we have tests that check whether the status is set to `available`?

pkg/resource/acl/hooks.go
kyriechen96: Modified. We have `synced` in `generator.yaml` file to set `ResourceSynced` to `true` if resource status is `available`

kyriechen96:
> Just one nit but otherwise a good catch. Do we have tests that check whether the status is set to `available`?

Ideally it doesn't need to be test, because we only set ResourceSynced to true when status is "available". It is a part in `generator.yaml` file. We can add a check in the existing yaml test files, but it might be a little tricky. Since controller doesn't validate AWS resource under the current test framework. We can only validate status of k8s resource. Also, acl, and user don't have yaml test file to validate it. Do you want me add test for those two resources as well?

pkg/resource/acl/hooks.go
```diff
@@ -45,8 +45,8 @@ func (rm *resourceManager) validateACLNeedsUpdate(
 	return nil
 }
 
-// aclActive returns true when the status of the given ACL is set to `active`
-func (rm *resourceManager) aclActive(
+// isACLActive returns true when the status of the given ACL is set to `active`
```
jljaco, line 48: you haven't mentioned this change in your PR description
if this and other changes are due to using a newer version of `code-generator`, then just state that in the PR description.

pkg/resource/cluster/hooks.go
```diff
@@ -265,12 +265,12 @@ func (rm *resourceManager) newMemoryDBClusterUploadPayload(
 	return res
 }
 
-// clusterActive returns true when the status of the given Cluster is set to `active`
-func (rm *resourceManager) clusterActive(
+// isClusterAvailable returns true when the status of the given Cluster is set to `active`
```
jljaco, line 268: you haven't updated this code comment with the new value

pkg/resource/snapshot/hooks.go
```diff
@@ -252,12 +252,12 @@ func (rm *resourceManager) newCopySnapshotPayload(
 	return res, nil
 }
 
-// snapshotActive returns true when the status of the given Snapshot is set to `active`
-func (rm *resourceManager) snapshotActive(
+// isSnapshotAvailable returns true when the status of the given Snapshot is set to `active`
```
jljaco, line 255: you haven't updated this code comment

jljaco (commented):
Looking good.  Would you please update your PR description to use proper formatting to make it clear when you're referring to fields and values?

Example:  

"`active` to `available`"

instead of 

"active to available"

pkg/resource/acl/hooks.go
kyriechen96: Added them into PR description~

pkg/resource/cluster/hooks.go
kyriechen96: Fixed

pkg/resource/snapshot/hooks.go
kyriechen96: Fixed

test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml
```diff
@@ -53,12 +53,12 @@ steps:
...
-#        - Key: "test_key_1"
-#          Value: "test_value_1"
-#        - Key: "test_key_2"
-#          Value: "test_value_2"
+    expect_aws:
+      Tags:
+        - Key: "test_key_1"
+          Value: "test_value_1"
+        - Key: "test_key_2"
+          Value: "test_value_2"
```
A-Hilaly, line 61: How is this asserting that a cluster is created?

kyriechen96: This is only asserting tags field of a cluster. Cluster creation is validated (assert) at step 1 in this yaml test file.
```
    expect_k8s:
      spec:
        name: cluster$RANDOM_SUFFIX
    expect_aws:
      Name: cluster$RANDOM_SUFFIX

File: test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml [hunk 6]
```diff
@@ -53,12 +53,12 @@ steps:
             value: "test_value_1"
           - key: "test_key_2"
             value: "test_value_2"
```
RedbackThomson, line 55: nit: Asserting `k8s` tags are on the spec isn't necessary. We know they are on the spec because lines 38-42 patched them onto it.

RedbackThomson (approved):

test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml
kyriechen96: How about other fields? If we do patch some fields, is it also not necessary to assert corresponding k8s fields?

A-Hilaly (approved):
Noice! :ship: it

test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml
RedbackThomson: The only time a spec field would change is if there is what we call "late initialisation" - the memorydb service changes the value, or provides a default value. In those cases it would be good to test that the service updated it as we would expect, and the spec now contains those values. Otherwise I think it's not necessary to test

RedbackThomson: That seems reasonable to me

jljaco (approved):
/lgtm

well done!

PR #57, opened 2023-02-07
Merge snapshot tests, and fix unreasonable timeout values.

kyriechen96:
Issue #, if available:
Snapshot tests creates 3 clusters. Reduce cluster creations to 2.
Some timeout values are not reasonable.

Description of changes:
Merge snapshot_validate_tags.yaml into other snapshot test.
Reduce timeout value of terminal cases to 60.
Reduce timeout value of tags update to 10.
Reduce timeout value of some huge wait time.


test/e2e/scenarios/ACL/acl_update_with_tags.yaml
```diff
@@ -60,7 +60,7 @@ steps:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 100
+            timeout: 10
```
RedbackThomson, line 63: Hmmm. Not related to this PRs changes, but this test only tests half of the output. To test the controllers we not only want to test that the K8s resource reached `ACK.ResourceSynced` but we also want to describe the AWS resource and make sure the tags appear on there too. We aren't doing any validation in any of these tests?

test/e2e/scenarios/Cluster/cluster_scale_in_scale_up_increase_replica.yaml
```diff
@@ -28,7 +28,7 @@ steps:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 7200
+            timeout: 5400
```
RedbackThomson, line 31: Could you add to the `description` above why these timeouts are so long? Explain about the snapshot window problem and why we can't just change the window to get around it

test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml
```diff
@@ -21,7 +21,7 @@ steps:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 7200
+            timeout: 3600
```
RedbackThomson, line 24: It takes an hour to start a memorydb cluster?

RedbackThomson (changes requested):
Left comments inline.
I think we could also reduce the wait timeout for terminal conditions to 10 - we would expect those to be fast because they are calling an AWS endpoint and parsing an error.

test/e2e/scenarios/Cluster/cluster_scale_in_scale_up_increase_replica.yaml
kyriechen96: Because we never know when we run e2e tests ...

kyriechen96: Yes we can change snapshot window to any time, but if we run e2e tests in that specific time range, then it is going to automatically snapshot.

kyriechen96: It is for all clusters, so I guess we don't need to add description for each one?

RedbackThomson: Why can't we make the snapshot window a dynamic variable? That way we can set it to be 12 hours apart from the time we start the tests?

test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml
kyriechen96: Same case here. Cluster creation takes less than 30 minutes. If the time is within Snapshot Window, it is going to snapshot cluster. Snapshotting should be done within 30 minutes as well. Hence, I set 1 hour in total to handle this case.

test/e2e/scenarios/ACL/acl_update_with_tags.yaml
kyriechen96: Should we add expect there? To be honestly I don't think our test framework supports Expect compare for Spec. Also, it added unit test for custom compare tags functions. Could the unit test handle this case?

test/e2e/scenarios/Cluster/cluster_scale_in_scale_up_increase_replica.yaml
kyriechen96: Do we have a way to set a dynamic variable in yaml file? Actually I don't know how to do it. Also, by default, If we do not specify this parameter, MemoryDB automatically chooses an appropriate time range.

test/e2e/scenarios/ACL/acl_update_with_tags.yaml
kyriechen96: Will fix that issue by a new PR to change test framework

test/e2e/scenarios/Cluster/cluster_scale_in_scale_up_increase_replica.yaml
kyriechen96: Will create a new PR to fix Snapshot Window issue.

kyriechen96: Update: `SnapshotWindow` issue is fixed in PR #66. Ideally cluster should finish creation within 30 minutes. However, I find that k8s needs to spend some time to finish `synced` after operations of a resource is done in aws service. In this case, it is rarely that k8s couldn't finish `synced` within 30 minutes after a cluster starts creating. Hence, I set all wait time for cluster and snapshot to 3600s to make sure they have enough time.

test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml
kyriechen96: Same as above.

A-Hilaly (approved):

RedbackThomson (approved):

RedbackThomson (approved):

test/e2e/scenarios/Cluster/cluster_update_with_tags.yaml
kyriechen96: https://github.com/aws-controllers-k8s/memorydb-controller/pull/57#discussion_r1125143899

test/e2e/scenarios/Cluster/cluster_scale_in_scale_up_increase_replica.yaml
RedbackThomson: This will be because we do an exponential backoff while waiting for updates. Our backoff window might grow past 30 minutes. 3600s is fine for that, then

PR #58, opened 2023-02-07
Ignore ClusterName for copySnapshot API

kyriechen96:
Issue #, if available:
v0.23.1 code-controller considers ClusterName as required field for snapshot, but memorydb copySnapshot API should have empty ClusterName.

Description of changes:
Set is_required value of ClusterName to false for snapshot.

### DIFF
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -209,6 +209,7 @@ resources:
           path: SourceSnapshotName
       ClusterName:
         is_primary_key: false
+        is_required: false
         from:
           operation: CreateSnapshot
           path: ClusterName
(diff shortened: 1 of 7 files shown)


apis/v1alpha1/generator.yaml
```diff
@@ -209,6 +209,7 @@ resources:
           path: SourceSnapshotName
       ClusterName:
         is_primary_key: false
+        is_required: false
```
RedbackThomson, line 212: I don't see any changes to the API file which I would expect to see if this worked

kyriechen96: Hopefully it could work

RedbackThomson: I think I understand - the `v0.23.1` code-generator fixed a bug where we weren't properly setting `required` on fields that were imported as custom. Because we now are, and this field was required, it was marking it as such even though we don't want it to be. So setting the `is_required` manually suppresses that. 👍🏼

PR #60, opened 2023-02-08
Add patching objects with tags tests

kyriechen96:
Issue #, if available:
Need to validate that tags could be added together with modification of other fields.

Description of changes:
Add changes of tags together while modifying other fields.
Remove a redundant test step.

### DIFF
--- a/test/e2e/scenarios/ACL/acl_create_update.yaml
+++ b/test/e2e/scenarios/ACL/acl_create_update.yaml
@@ -85,8 +85,8 @@ steps:
       UserNames:
         - userone$RANDOM_SUFFIX
         - usertwo$RANDOM_SUFFIX
-  - id: "UPDATE_ACL"
-    description: "Update userNames"
+  - id: "UPDATE_ACL_WITH_USERNAMES_AND_TAGS"
+    description: "Update userNames and tags"
     patch:
       apiVersion: $CRD_GROUP/$CRD_VERSION
       kind: ACL
@@ -95,6 +95,9 @@ steps:
       spec:
         userNames:
           - userone$RANDOM_SUFFIX
+        tags:
+          - key: "test_key_1"
+            value: "test_value_1"
     wait:
       status:
         conditions:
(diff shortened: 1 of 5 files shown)


test/e2e/scenarios/ACL/acl_create_update.yaml
```diff
@@ -74,12 +74,15 @@ steps:
       spec:
         userNames:
           - userone$RANDOM_SUFFIX
+        tags:
+          - key: "test_key_1"
+            value: "test_value_1"
```
A-Hilaly, line 79: Curious here, how are you asserting that these tags got applied to the ACL?

RedbackThomson: I had the same comment in another PR - we aren't validating that the AWS resource actually has the tags.

test/e2e/scenarios/ACL/acl_create_update.yaml
```diff
@@ -74,12 +74,15 @@ steps:
...
     wait:
       status:
         conditions:
           ACK.ResourceSynced:
             status: "True"
-            timeout: 180
+            timeout: 300
```
RedbackThomson, line 85: No way it takes 5 minutes to update tags. That's insane!

kyriechen96: With 180 second it failed. I checked controller logs, and logs show updating tags all the time (with different orders) so `ResourceSynced` cannot be true. Should we pause this PR and wait for new tag compare function being added to `code-generator` and `runtime`?

kyriechen96: I've tried locally. It really took a long time to finish update. More than 5 minutes. Really think we need to wait for new version of code-generator/runtime that includes new compare tags function/method. Extra tags update calls slow everything down.

test/e2e/scenarios/ACL/acl_create_update.yaml
kyriechen96: Resolved in PR #65

A-Hilaly (approved):
+++

RedbackThomson (approved):
/lgtm

PR #65, opened 2023-02-20
Assert expected spec with kubernetes resources and all fields of aws resources

kyriechen96:
Issue #, if available:
Current test framework only checks whether ResourceSynced is true or not for Kubernetes resource. It doesn't check spec of Kubernetes resource, and it doesn't check aws resource.

Description of changes:
1. Add a comparison between expectations and Spec of Kubernetes resources to assert.
2. Add a comparison between expectations and MemoryDB resources to assert.
3. Modify test files to fit new test framework.
4. Fix issue 1 and 2.
5. Comment issue 3 and will create a new PR to investigate.

Note: Add custom logic of comparison to tags, parameterNameValues, clusterName, subnetIDs. New custom logic also shows details if assertError happens, which is easier for debugging.

Issues caught by latest test framework.
1. MemoryDB doesn't support parameter group creation with parameterNameValues change.
2. Controller only reset parameters in parameter group to default when desired parameters have empty value. Controller doesn't update values of parameters.
3. Tags for cluster and snapshot cannot be updated.


test/e2e/tests/test_scenarios.py
```diff
@@ -37,6 +37,63 @@ class ParameterGroupHelper(helper.ResourceHelper):
...
+            # there are expectations but no given state to validate
+            # following assert will fail and assert introspection will provide useful information for debugging
+            assert expectations == state
+
+        for (key, value) in expectations.items():
+            # conditions are processed separately
+            if key == "conditions":
+                continue
+            # tags are processed by custom logic
+            if key == "tags":
```
RedbackThomson, line 63: Rather than supporting each key individually, we could do a comparison of value to state, based on the key?
Something like the following:
```python
for (key, value) in expectations.items():
  if key not in expectations:
    continue
  assert state.get(key) == value
```

I think this should work for most of the basic string-string comparisons. Possibly for tags it wouldn't work, but we could also deal with those by sorting and then comparing.

test/e2e/scenarios/ACL/acl_create_update.yaml
```diff
@@ -80,6 +80,10 @@ steps:
           ACK.ResourceSynced:
             status: "True"
             timeout: 180
+    expect:
```
RedbackThomson, line 83: We will want assertions for AWS after every step. After the create, check the resource was actually created in memorydb. Check the important parts of the spec match the created resource. Especially check that any part of the spec that uses custom code was created/updated properly.

test/e2e/scenarios/ACL/acl_update_with_tags.yaml
```diff
@@ -54,13 +54,20 @@ steps:
...
           ACK.ResourceSynced:
             status: "True"
             timeout: 100
+    expect:
+      spec:
+        tags:
+          - key: "test_key_1"
+            value: "test_value_1"
+          - key: "test_key_2"
+            value: "test_value_2"
```
RedbackThomson, line 70: Looks like this new solution already found a bug!

RedbackThomson (changes requested):

test/e2e/tests/test_scenarios.py
kyriechen96: Yes, that is what we do. For most cases, we just do `assert (key, value) == (key, state.get(key))`. For `conditions`, `tags`, and `parameterNameValues`, we do check(assert) something different.

test/e2e/scenarios/ACL/acl_create_update.yaml
kyriechen96: I think if a resource is synced which means the resource exists and is created successfully. Correct me if I'm wrong. I can add spec check for resource creations if you think that's necessary~

test/e2e/tests/test_scenarios.py
```diff
@@ -37,6 +37,63 @@ class ParameterGroupHelper(helper.ResourceHelper):
     Overrides methods as required for custom resources.
     """
 
+    def assert_expectations(self, verb: str, input_data: dict, expectations: model.ExpectDict, reference: k8s.CustomResourceReference) -> None:
```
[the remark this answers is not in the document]
kyriechen96: I've talked with Kumar for this issue before when I added custom logic for checking Events. He asked me to add them separately for each resource, and it is easier to maintain the test framework (for cases a part of resources have error).

test/e2e/scenarios/ACL/acl_update_with_tags.yaml
kyriechen96: Actually this is not a bug. I checked MemoryDB APIs, and MemoryDB doesn't support a null key for tags (ideally each tag should have both key and value). For non-null key with null value case, MemoryDB automatically fills a null value for the key. I've just added value for this case so that it won't confuse others.

test/e2e/tests/test_scenarios.py
```diff
@@ -37,6 +37,63 @@ class ParameterGroupHelper(helper.ResourceHelper):
...
+       logging.debug(f"ParameterGroup - assert_expectations()")
+       self._assert_conditions(expectations, reference, wait=False)
+       # conditions expectations met, now check current resource against expectations
+       resource = k8s.get_resource(reference)
+       self.assert_items(expectations.get("status"), resource.get("status"))
+       self.assert_items(expectations.get("spec"), resource.get("spec"))
```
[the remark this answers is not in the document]
kyriechen96: I will add another function to assert with boto3 resources.

test/e2e/tests/test_scenarios.py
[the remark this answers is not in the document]
kyriechen96: Integrated them into parent function

test/e2e/scenarios/ACL/acl_create_update.yaml
kyriechen96: Added.

test/e2e/tests/test_scenarios.py
```diff
@@ -90,6 +283,61 @@ class ClusterHelper(helper.ResourceHelper):
...
+
+        for (key, value) in expectations.items():
+            # describe_clusters API provides Name, Status, NumberOfShards, ClusterEndpoint,
+            # NodeType, EngineVersion, EnginePatchVersion, ParameterGroupName, ParameterGroupStatus,
+            # SecurityGroups, SubnetGroupName, TLSEnabled, ARN, SnapshotRetentionLimit,
+            # MaintenanceWindow, SnapshotWindow, ACLName, AutoMinorVersionUpgrade.
+            # Currently we only validate Name, NodeType, ACLName, NumShards, SnapshotWindow
+            # SnapshotRetentionLimit, MaintenanceWindow, ParameterGroupName
+            # NodeType is processed by custom logic
+            if key == "NodeType":
```
RedbackThomson, line 311: I don't understand why we need to map the key to a boto3 response path. Why not just require that the key be the field we want to assert within the response? 
```yaml
assert:
  ACLName: <value>
```

and then for nested fields within the response, we just nest the assertion:
```yaml
assert:
  Shards:
    NumberOfNodes: <value>
```

test/e2e/tests/test_scenarios.py
```diff
@@ -37,6 +37,88 @@ class ParameterGroupHelper(helper.ResourceHelper):
...
+    def assert_items_boto3(self, expectations: dict, state: dict) -> None:
+        if not expectations:
+            # nothing to assert as there are no expectations
+            return
+        if not state:
+            # there are expectations but no given state to validate
+            # following assert will fail and assert introspection will provide useful information for debugging
+            assert expectations == state
+
+        mdb = boto3.client("memorydb")
```
RedbackThomson, line 97: You can put this line in the constructor of `ResourceHelper` and store it as a member variable, that way we aren't initialising it for every `assert` call.

test/e2e/tests/test_scenarios.py
```diff
@@ -37,6 +37,88 @@ class ParameterGroupHelper(helper.ResourceHelper):
     Overrides methods as required for custom resources.
     """
 
+    def assert_items_k8s(self, expectations: dict, state: dict) -> None:
```
RedbackThomson, line 40: Rather than overwriting the entire `assert_items_k8s` method, we should use the common logic that is defined for it in `ResourceHelper`. Then, when it gets to the custom logic inside the `for` loop, it could instead call another method which we overwrite. For example:

In `helper.py`:
```python
    def assert_items_k8s(self, expectations: dict, actual: dict) -> None:
        """Asserts the values in the `expectations` dict match the values as
        provided in the actual `dict`. This method calls into `assert_extra_k8s`
        for any items that can't be asserted in general.

        Args:
            expectations: dictionary with items (expected) to assert
            actual: dictionary of the latest spec or state of the K8s resource
        Returns:
            None
        """

        if not expectations:
            # nothing to assert as there are no expectations
            return

        if not actual:
            # there are expectations but no given state to validate
            # following assert will fail and assert introspection will
            # provide useful information for debugging
            assert expectations == actual

        for (key, value) in expectations.items():
            # conditions are asserted separately
            if key == "conditions":
                continue

            # tags are processed by custom logic
            if key == "tags":
                assertResourceTags(value, actual.get(key))
                continue

            self.assert_extra_k8s(key, value, actual)

            assert value == actual.get(key)

    def assert_extra_k8s(self, key: str, value: dict, actual: dict):
        # This method should be overridden by the any implementations of
        # the `ResourceHelper` that need to assert additional custom properties
        pass

    def assertResourceTags(self, expectedTags: dict, actualTags: dict):
        for expectedTag in expectedTags:
            tagKeyExist = False
            tagKey = expectedTag.get("key")

            for actualTag in actualTags:
                if ("key", tagKey) in actualTag.items():
                    assert expectedTag == actualTag
                    tagKeyExist = True
                    break

            if tagKeyExist:
                continue
```

Then in your individual implementations of `ResourceHelper`, extend it only the custom logic:
```python
    def assert_extra_k8s(self, key: str, value: dict, actual: dict):
        # This method should be overridden by the any implementations of
        # the `ResourceHelper` that need to assert additional custom properties
        if key == "parameterNameValues":
            if not value:
                assert value == actual.get(key)
                return

            for expectedParameterNameValue in value:
                parameterNameExist = False
                parameterName = expectedParameterNameValue.get("parameterName")
                latestParameterNameValues = actual.get(key)
                for latestParameterNameValue in latestParameterNameValues:
                    if ("parameterName", parameterName) in latestParameterNameValue.items():
                        assert expectedParameterNameValue == latestParameterNameValue
                        parameterNameExist = True
                        break
                if parameterNameExist:
                    continue
                assert expectedParameterNameValue == None
                
            return
```

RedbackThomson: You should follow this same pattern for the boto3 style. It should greatly reduce code repetition and improve readability

test/e2e/tests/test_scenarios.py
```diff
@@ -37,6 +37,88 @@ class ParameterGroupHelper(helper.ResourceHelper):
...
+
+        mdb = boto3.client("memorydb")
+        parameterGroupName = ""
+        for (latestKey, latestValue) in state.items():
+            if latestKey == "name":
+                parameterGroupName = latestValue
+        parameterGroups = mdb.describe_parameter_groups(ParameterGroupName=parameterGroupName)
+        parameterGroup = parameterGroups['ParameterGroups'][0]
+        latestTags = mdb.list_tags(ResourceArn=parameterGroup.get("ARN"))['TagList']
+        parameterNameValues = mdb.describe_parameters(ParameterGroupName=parameterGroupName)['Parameters']
```
RedbackThomson, line 105: This should be moved into a method that each individual `ResourceHelper` overrides, then call it from within `ResourceHelper.assert_expectations` and pass the result as the `state` (please rename this to `actual`) into `assert_items_boto3`

RedbackThomson (changes requested):

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -122,9 +122,11 @@ def assert_expectations(self, verb: str, input_data: dict, expectations: model.E
...
         # conditions expectations met, now check current resource against expectations
         resource = k8s.get_resource(reference)
-        self.assert_items(expectations.get("status"), resource.get("status"))
-
-        # self._assert_state(expectations.get("spec"), resource)  # uncomment to support spec assertions
+        # status and spec assertions with resources from Kubernetes
+        self.assert_items_k8s(expectations.get("status"), resource.get("status"))
+        self.assert_items_k8s(expectations.get("spec"), resource.get("spec"))
+        # spec assertions with resources from boto3
+        self.assert_items_boto3(expectations.get("spec"), resource.get("spec"))
```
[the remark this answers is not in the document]
kyriechen96: boto3 API calls don't have field as "spec" or "status". I'm creating a new class only for boto3 now

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -186,7 +188,27 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
...
                 raise Exception(f"Condition {condition_name} is provided with invalid value: {expected_value} ")
 
-    def assert_items(self, expectations: dict, state: dict) -> None:
+    def assert_items_boto3(self, expectations: dict, state: dict) -> None:
+            """Asserts boto3 against supplied expectations
+            Override it as needed for custom verifications
+
+            Args:
+                expectations: dictionary with items (expected) to assert in state
+                state: dictionary with items (actual)
```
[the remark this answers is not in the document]
kyriechen96: state here is still from k8s. Needs to get resource name from reference first, and then call boto3 later.

test/e2e/tests/test_scenarios.py
kyriechen96: Creating a new class for boto3 now, so boto3 resources validation doesn't need to do together with k8s resources.

test/e2e/tests/test_scenarios.py
kyriechen96: Modified.

test/e2e/tests/test_scenarios.py
kyriechen96: This way will miss a `continue` after calling `assert_extra_k8s`. I moved them to helper.py

test/e2e/tests/test_scenarios.py
kyriechen96: I don't find a way to override a part in for loop that calls continue/break there. Same as last comment.

kyriechen96: Mostly need to override entire `assert_items_boto3` function.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -186,13 +203,32 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
             else:
                 raise Exception(f"Condition {condition_name} is provided with invalid value: {expected_value} ")
 
-    def assert_items(self, expectations: dict, state: dict) -> None:
-        """Asserts state against supplied expectations
+    def assert_items_boto3(self, expectations: dict, actual: dict) -> None:
+        """Asserts actual against supplied expectations
+        Override it as needed for custom verifications
```
A-Hilaly, line 206: I don't see any override operations in this function

test/e2e/tests/test_scenarios.py
```diff
@@ -128,3 +308,39 @@ class TestScenarios:
...
+            if expectedParameter == latestParameter:
+                parameterExist = True
+                break
+        if parameterExist:
+            continue
+        assert expectedParameter == None
+
+# assertSubnets returns AssertError if subnets don't match
+def assertSubnets(expectedSubnets, latestSubnets):
+    for expectedSubnet in value:
```
A-Hilaly, line 344: Did you mean `for expectedSubnet in expectedSubnets` ?

test/e2e/tests/test_scenarios.py
```diff
@@ -128,3 +308,39 @@ class TestScenarios:
...
+def assertTags(expectedTags, latestTags):
+    for expectedTag in expectedTags:
+        tagKeyExist = False
+        for latestTag in latestTags:
+            if expectedTag == latestTag:
+                tagKeyExist = True
+                break
+        if tagKeyExist:
+            continue
+        assert expectedTag == None
```
A-Hilaly, line 328: Maybe
```python
import collections
collections.Counter(x) == collections.Counter(y)
```

test/e2e/tests/test_scenarios.py
```diff
@@ -90,6 +239,37 @@ class ClusterHelper(helper.ResourceHelper):
...
+            # there are expectations but no given actual state to validate
+            # following assert will fail and assert introspection will provide useful information for debugging
+            assert expectations == actual
+
+        clusterName = ""
+        for (latestKey, latestValue) in actual.items():
+            if latestKey == "name":
+                clusterName = latestValue
+        clusters = mdb.describe_clusters(ClusterName=clusterName)
+        cluster = clusters['Clusters'][0]
```
A-Hilaly, line 262: QQ: what if the cluster didn't get created? Is it correct to assume that `describe_clusters` will always return a non empty earray?

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -201,16 +235,50 @@ def assert_items(self, expectations: dict, state: dict) -> None:
...
+                    latestTags = actual.get(key)
+                    for latestTag in latestTags:
+                        if ("key", tagKey) in latestTag.items():
+                            assert expectedTag == latestTag
+                            tagKeyExist = True
+                            break
+                    if tagKeyExist:
+                        continue
+                    assert expectedTag == None
+                continue
```
A-Hilaly, line 261: How about the tag value?

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +112,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
         sleep(self.DEFAULT_WAIT_SECS)
         self.wait_for_delete(reference)  # throws exception if wait fails
 
-    def assert_expectations(self, verb: str, input_data: dict, expectations: model.ExpectDict, reference: k8s.CustomResourceReference) -> None:
+    def assert_expectations_k8s(self, verb: str, input_data: dict, expectations_k8s: model.ExpectDictK8S, reference: k8s.CustomResourceReference) -> None:
```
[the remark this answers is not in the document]
kyriechen96: Previous test framework used `expectations`. I guess the reason is that we have more than one expected spec/status fields to compare in yaml test file. Do you think I should change it to the `expectation`?

test/e2e/declarative_test_fwk/helper.py
kyriechen96: I override this function for each resource in `test_scenarios.py` file

test/e2e/tests/test_scenarios.py
kyriechen96: Oh yeah, typo there

test/e2e/tests/test_scenarios.py
kyriechen96: expectTags is from manual input in yaml test file. In some cases, it might only list the part of modified tags instead of all tags (otherwise yaml test file may look too messy).

test/e2e/declarative_test_fwk/helper.py
kyriechen96: expectedTag and latestTag include both tag key and tag value

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +110,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
...
+        :param input_data: input data to verb
+        :param expectations_boto3: expectations to assert
+        :param reference: custom resource reference
+        :return: None
+        """
+        resource = k8s.get_resource(reference)
+        # assertions with resources from boto3
+        # spec of k8s resource here is for getting name of resource
+        # boto3 call APIs using name of resource
+        self.assert_items_boto3(expectations_boto3, resource.get("spec"))
```
RedbackThomson, line 142: This line implies that we can only assert that the spec fields of our resource match the response given by boto3. There are times where we may want to check that the status fields, too. For example, I know we have custom code for adding `events` into the status - we should check for that here too? Make sure that the # of events in the status matches the # of events returned by boto3

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -201,16 +235,50 @@ def assert_items(self, expectations: dict, state: dict) -> None:
...
+                        if ("key", tagKey) in latestTag.items():
+                            assert expectedTag == latestTag
+                            tagKeyExist = True
+                            break
+                    if tagKeyExist:
+                        continue
+                    assert expectedTag == None
+                continue
+            # parameterNameValues are processed by custom logic
+            if key == "parameterNameValues":
```
RedbackThomson, line 263: Should this method be in the `ResourceHelper` class? This condition is only valid for the `ParameterGroup` implementation?

test/e2e/declarative_test_fwk/model.py
```diff
@@ -54,10 +54,13 @@ class DeleteDict(ResourceDict):
...
 
 # fields for 'expect' field in a test step
-class ExpectDict(TypedDict, total=False):
+class ExpectDictK8S(TypedDict, total=False):
     spec: Dict
     status: Dict
 
+class ExpectDictBOTO3(TypedDict, total=False):
+    spec: Dict
+    status: Dict
```
RedbackThomson, line 63: There is no such thing as a `spec` or `status` in the response of a boto3 resource. Instead, on line #74, we could just make it `expect_boto3: Dict`

RedbackThomson (changes requested):

test/e2e/declarative_test_fwk/helper.py
kyriechen96: `resource.get("spec")` here is only used for getting `name` of resources. Then we call APIs from boto3 by using `name` of resources. API returns both `spec` and `status`. Hence, it already covers `status`

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Currently this condition is only valid for `ParameterGroup`. Other resources don't have this field, so they will skip this condition. However, describe_parameters is a independent API which is not related to describe_<resource type> APIs, so I put it into `ResourceHelper` class.

test/e2e/declarative_test_fwk/model.py
kyriechen96: Got it

test/e2e/tests/test_scenarios.py
kyriechen96: I've tried in python and function can pass parameters. However, when I tried to put `mdb` into `ResourceHelper`, it reports an error in other-resourceHelp(ResourceHelper) that `mdb is not defined`. Hence, I will keep this way.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +112,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
...
+        self._assert_conditions(expectations_k8s, reference, wait=False)
+        # conditions expectations met, now check current resource against expectations_k8s
         resource = k8s.get_resource(reference)
-        self.assert_items(expectations.get("status"), resource.get("status"))
+        # status and spec assertions with resources from Kubernetes
+        self.assert_items_k8s(expectations_k8s.get("status"), resource.get("status"))
+        self.assert_items_k8s(expectations_k8s.get("spec"), resource.get("spec"))
 
-        # self._assert_state(expectations.get("spec"), resource)  # uncomment to support spec assertions
+    def assert_expectations_boto3(self, verb: str, input_data: dict, expectations_boto3: model.ExpectDictBOTO3, reference: k8s.CustomResourceReference) -> None:
```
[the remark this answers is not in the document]
kyriechen96: I will modify to a proper name

test/e2e/tests/test_scenarios.py
kyriechen96: If cluster didn't get created, it will return exception `MemoryDB.Client.exceptions.ClusterNotFoundFault` at line 255.

test/e2e/declarative_test_fwk/helper.py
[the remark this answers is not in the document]
kyriechen96: The name here I use is for asserting expectations from k8s resource. Also another is for asserting expectations from aws resource.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +110,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
...
+        # conditions expectations met, now check current resource against expectations_k8s
         resource = k8s.get_resource(reference)
-        self.assert_items(expectations.get("status"), resource.get("status"))
+        # status and spec assertions with resources from Kubernetes
+        self.assert_items_k8s(expectations_k8s.get("status"), resource.get("status"))
+        self.assert_items_k8s(expectations_k8s.get("spec"), resource.get("spec"))
 
-        # self._assert_state(expectations.get("spec"), resource)  # uncomment to support spec assertions
+    def assert_api_response(self, verb: str, input_data: dict, expectations_aws: model.ExpectAPIResponseDict, reference: k8s.CustomResourceReference) -> None:
+        """Asserts custom resource reference in AWS MemoryDB service against the supplied expectations
```
RedbackThomson, line 130:
```suggestion
"""Asserts that the given set of expectations matches the response provided by the MemoryDB service
```

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -186,13 +201,32 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
             else:
                 raise Exception(f"Condition {condition_name} is provided with invalid value: {expected_value} ")
 
-    def assert_items(self, expectations: dict, state: dict) -> None:
-        """Asserts state against supplied expectations
+    def assert_items_aws(self, expectations: dict, actual: dict) -> None:
+        """Asserts actual against supplied expectations
         Override it as needed for custom verifications
```
RedbackThomson, line 206: `assert_items_aws` does not `actual` against the supplied expectations. It asserts the boto3 response against the supplied expectations.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -186,13 +201,32 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
...
 
         Args:
-            expectations: dictionary with items (expected) to assert in state
-            state: dictionary with items (actual)
+            expectations: dictionary with items (expected) to assert in actual state
+            actual: dictionary with items (actual)
```
RedbackThomson, line 210:
```suggestion
actual: the current spec of the K8s resource
```

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -186,13 +201,32 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
             else:
                 raise Exception(f"Condition {condition_name} is provided with invalid value: {expected_value} ")
 
-    def assert_items(self, expectations: dict, state: dict) -> None:
-        """Asserts state against supplied expectations
+    def assert_items_aws(self, expectations: dict, actual: dict) -> None:
```
RedbackThomson, line 204: I would suggest changing `actual` to just `spec` here. We use the terms `expect` and `actual` to refer to the things being compared. But this method doesn't compare `actual`, it compares the boto3 response.

test/e2e/declarative_test_fwk/helper.py
RedbackThomson: If this condition is only valid for `ParameterGroup`, then it should live in the helper for that class. If I was trying to debug the code for the `ParameterGroup` tests, I wouldn't think to look here for that logic.

test/e2e/declarative_test_fwk/model.py
```diff
@@ -53,11 +53,14 @@ class DeleteDict(ResourceDict):
...
-# fields for 'expect' field in a test step
-class ExpectDict(TypedDict, total=False):
+# fields for 'expect_k8s' field in a test step
+class ExpectK8SDict(TypedDict, total=False):
     spec: Dict
     status: Dict
 
+#fields for 'expect_aws' field in a test step
+class ExpectAPIResponseDict(TypedDict, total=False):
+    expect_aws: Dict
```
RedbackThomson, line 63: This is going to lead to a YAML that looks like:
```yaml
...
expect_aws:
  expect_aws:
    ...
```

Which isn't necessary. I don't think you need to define a type `ExpectAPIResponseDict`. Just, only line 74, do `expect_aws: Dict`. It doesn't need a wrapper class.

test/e2e/declarative_test_fwk/runner.py
```diff
@@ -193,17 +194,38 @@ def assert_expectations(step: model.Step) -> None:
...
+    """Asserts api_response as specified in the Step.
+
+    Args:
+        step: test step
+
+    Returns:
+        None
+    """
+
+    logging.info(f"assert:  {step}")
```
RedbackThomson, line 218: Is this logging meant to be here?

test/e2e/tests/test_scenarios.py
```diff
@@ -37,6 +37,39 @@ class ParameterGroupHelper(helper.ResourceHelper):
...
     """
 
+    def assert_items_aws(self, expectations: dict, actual: dict) -> None:
+        if not expectations:
+            # nothing to assert as there are no expectations
+            return
+        if not actual:
+            # there are expectations but no given actual state to validate
+            # following assert will fail and assert introspection will provide useful information for debugging
+            assert expectations == actual
```
RedbackThomson, line 47: As per my previous, very long, answer - https://github.com/aws-controllers-k8s/memorydb-controller/pull/65#discussion_r1116301917

I don't believe that this should be completely overriding `assert_items_aws`. I think it's a bad OOP pattern, when we could be using smaller, more well defined, methods.

test/e2e/tests/test_scenarios.py
RedbackThomson: I am not satisfied with that answer. I am not talking about functions and parameters, I am talking about initialising a member variable inside the constructor of the parent class.

test/e2e/scenarios/Cluster/cluster_create_update.yaml
```diff
@@ -29,6 +29,28 @@ steps:
...
+        tlsEnabled: true
+        snapshotWindow: 05:00-06:00
+        snapshotRetentionLimit: 0
+    expect_aws:
+      Name: cluster$RANDOM_SUFFIX
+      NodeType: db.t4g.medium
+      ACLName: open-access
+      NumberOfShards: 1
+      Description: Test create from ACK
+      SnsTopicArn: $TOPIC1
```
nmvk, line 50: We need to assert status entries?

test/e2e/scenarios/Cluster/cluster_scale_in_scale_up_increase_replica.yaml
```diff
@@ -46,6 +56,14 @@ steps:
...
             status: "True"
             timeout: 10000
+    expect_k8s:
+      spec:
+        nodeType: db.t4g.medium
+        numShards: 1
+        numReplicasPerShard: 1
+    expect_aws:
+      NodeType: db.t4g.medium
+      NumberOfShards: 1
```
nmvk, line 66: nit: assert num replicas too

test/e2e/declarative_test_fwk/helper.py
kyriechen96: It not only compares `spec` but also compares `status`. API calls from boto3 response return fields from both `status` and `spec`. It doesn't really have keywords called `status` or `spec` from boto3 response.

test/e2e/tests/test_scenarios.py
```diff
@@ -128,3 +314,42 @@ class TestScenarios:
...
+def assertSubnets(expectedSubnets, latestSubnets):
+    for expectedSubnet in expectedSubnets:
+        subnetExist = False
+        for latestSubnet in latestSubnets:
+            if expectedSubnet.items() <= latestSubnet.items():
+                subnetExist = True
+                break
+        if subnetExist:
+            continue
+        assert expectedSubnet == None
```
nmvk, line 355: This isn't asserting the individual values?

nmvk (changes requested):

test/e2e/declarative_test_fwk/helper.py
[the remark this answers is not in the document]
kyriechen96: Modified to a clearer name

test/e2e/tests/test_scenarios.py
kyriechen96: Incorrectly set it to an instant variable. Put it into constructor now.

test/e2e/tests/test_scenarios.py
kyriechen96: Created a function to override and pass boolean value to `assert_items_k8s`

test/e2e/declarative_test_fwk/runner.py
```diff
@@ -193,17 +194,38 @@ def assert_expectations(step: model.Step) -> None:
...
+    if not step.expectations_aws:
```
jljaco, line 219: Why not `assert step.expectations_aws is not nil`?

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -201,16 +235,50 @@ def assert_items(self, expectations: dict, state: dict) -> None:
...
+            assert expectations == actual
 
         for (key, value) in expectations.items():
             # conditions are processed separately
             if key == "conditions":
                 continue
-            assert (key, value) == (key, state.get(key))
+            # tags are processed by custom logic
+            if key == "tags":
+                for expectedTag in value:
```
jljaco, line 249: Can this custom logic be implemented in a dedicated helper function that can be independently tested/validated?

test/e2e/declarative_test_fwk/runner.py
jljaco: Perhaps this should become `logging.debug`?

test/e2e/tests/test_scenarios.py
kyriechen96: Added a `get_aws_resource` function for it

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Added `assert_extra_items_k8s` function to handle it. Override the function in `ResourceHelper` for `ParameterGroup`

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Modified.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Changed all `actual` to `k8s_resource` to make code more readable.

test/e2e/declarative_test_fwk/model.py
kyriechen96: Updated to regular dict

test/e2e/declarative_test_fwk/runner.py
kyriechen96: I've changed this one to `logging.debug`. However, `runner.py file` has both `logging.info` and `logging.debug` inside. Should I change all `logging.info` to `logging.debug`? I feel like either one is fine, but I want to keep them consistent.

test/e2e/tests/test_scenarios.py
kyriechen96: Yeah added `assert_extra_items_aws` and `get_aws_resource` inside of `assert_items_aws`. Only override those two functions.

test/e2e/scenarios/ParameterGroup/pg_reset_params.yaml
```diff
@@ -1,41 +0,0 @@
-id: "PG_RESET_PARAMS"
-description: "In this test we create PG with Params and Reset them"
-#marks:
-#  - slow
-#  - blocked
-resource:
-  apiVersion: $CRD_GROUP/$CRD_VERSION
-  kind: ParameterGroup
```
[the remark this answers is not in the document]
kyriechen96: I merged `pg_reset_params.yaml` into `pg_update_with_params.yaml`. Steps in `pg_update_with_params.yaml` are literally same as required steps in `pg_reset_params.yaml`. Only one yaml test file could cover both validations.

test/e2e/scenarios/ACL/acl_terminal_condition.yaml
```diff
@@ -58,6 +63,15 @@ steps:
...
+        name: acl$RANDOM_SUFFIX
+        userNames:
+          - user$RANDOM_SUFFIX
+    expect_aws:
+      Name: acl$RANDOM_SUFFIX
+      UserNames:
+        - user$RANDOM_SUFFIX
   - id: "UPDATE_ACL_INVALID"
     description: "Update userNames to invalid"
     patch:
```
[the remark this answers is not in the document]
kyriechen96: Yeah, added.

test/e2e/scenarios/Cluster/cluster_create_update.yaml
kyriechen96: Do you mean we need to add status assertion? If so, what fields do you want me to add? Or do you mean why we assert `NumberOfShards` (from status) here? Because describe_clusters API called by boto3 doesn't have `NumShards`. It only has `NumberOfShards`, so that is the only field to validate number of shards.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/memorydb/client/describe_clusters.html

test/e2e/scenarios/Cluster/cluster_scale_in_scale_up_increase_replica.yaml
kyriechen96: I don't find a field related to number of replicas from `describe_clusters` API called by boto3.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/memorydb/client/describe_clusters.html

test/e2e/tests/test_scenarios.py
kyriechen96: I think it does. In `expectedSubnet`, there are only one field `Identifier` (name). In latestSubnet from `describe_subnet_groups` called by `boto3`, there are other fields except `Identifier`. We only need to check if names match.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/memorydb/client/describe_subnet_groups.html

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +110,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
         sleep(self.DEFAULT_WAIT_SECS)
         self.wait_for_delete(reference)  # throws exception if wait fails
 
-    def assert_expectations(self, verb: str, input_data: dict, expectations: model.ExpectDict, reference: k8s.CustomResourceReference) -> None:
+    def assert_expectations_k8s(self, verb: str, input_data: dict, expectations_k8s: model.ExpectK8SDict, reference: k8s.CustomResourceReference) -> None:
         """Asserts custom resource reference inside Kubernetes cluster against the supplied expectations
 
-        :param verb: expectations after performing the verb (apply, patch, delete)
+        :param verb: expectations_k8s after performing the verb (apply, patch, delete)
```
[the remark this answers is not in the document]
kyriechen96: It is from original test framework code, but I will change it.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +110,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
...
-        self.assert_items(expectations.get("status"), resource.get("status"))
+        # status and spec assertions with resources from Kubernetes
+        self.assert_items_k8s(expectations_k8s.get("status"), resource.get("status"))
+        self.assert_items_k8s(expectations_k8s.get("spec"), resource.get("spec"))
 
-        # self._assert_state(expectations.get("spec"), resource)  # uncomment to support spec assertions
+    def assert_api_response(self, verb: str, input_data: dict, expectations_aws: model.ExpectAPIResponseDict, reference: k8s.CustomResourceReference) -> None:
+        """Asserts custom resource reference in AWS MemoryDB service against the supplied expectations
+
+        :param verb: expectations_aws after performing the verb (apply, patch, delete)
```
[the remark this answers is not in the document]
kyriechen96: Modified.

test/e2e/declarative_test_fwk/runner.py
kyriechen96: Because `expecatations_aws` can be null. It is not a required field. For example, when we validate terminal condition from a resource creation, the resource is not created due to terminal exceptions. Hence, don't need e`xpectations_aws` field in this case.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Do you want me to add unit tests for those custom logic? There are 2 custom logic for validating k8s resource and 3 custom logic for validating aws resource.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Update: don't need to override this function anymore.

test/e2e/declarative_test_fwk/helper.py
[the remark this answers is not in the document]
kyriechen96: Changed them to `assert_k8s_resource` and `assert_aws_resource` to make them more readable.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -176,23 +195,85 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
...
+
+        Args:
+            mdb: memorydb client from boto3
+            aws_resource: specific resource from aws memorydb
+            k8s_resource: the current status/spec of the k8s resource
+            expected_name: expected name of a specific field to assert in aws resource
+            expected_value: expected value of a specific field to assert in aws resource
+
+        Returns:
+            True/False
```
RedbackThomson, line 267: nit: Explain why you would return true. eg. `Return true if the expected_name was asserted successfully, false if it was not asserted`

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -176,23 +195,85 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
...
+
+        Args:
+            k8s_resource: the current status/spec of the k8s resource
+
+        Returns:
+            aws resource
+        """
+        return
+
+    def assert_extra_items_aws(self, aws_resource, expected_name, expected_value):
```
RedbackThomson, line 255:
```suggestion
def assert_extra_items_aws(self, aws_resource, expected_name, expected_value) -> bool:
```

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -176,23 +195,85 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
...
+        """Asserts specific field boto3 response against supplied expectation
+        Override it as needed for validations of specific field
+
+        Args:
+            mdb: memorydb client from boto3
```
RedbackThomson, line 260: This arg is gone

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -176,23 +195,85 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
...
+            aws_resource: specific resource from aws memorydb
+            k8s_resource: the current status/spec of the k8s resource
```
RedbackThomson, line 262: This arg is gone too

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -201,16 +282,57 @@ def assert_items(self, expectations: dict, state: dict) -> None:
...
+            # tags are processed by custom logic
+            if key == "tags":
+                self.assert_tags(value, k8s_resource.get(key))
+                continue
+            # assert any specific extra fields
+            if self.assert_extra_items_k8s(key, value, k8s_resource):
+                continue
+            assert value == k8s_resource.get(key)
+
+    def assert_extra_items_k8s(self, name: str, value: any, k8s_resource: dict) -> bool:
```
RedbackThomson, line 303: nit: The parameters for this method have different names and order to `assert_extra_items_aws`

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -201,16 +282,57 @@ def assert_items(self, expectations: dict, state: dict) -> None:
...
+        """Asserts extra fields from k8s resource against fields from expectations
+        Override it as needed for custom verifications
+
+        Args:
+            name: specific field name of expectations
+            value: specific field value of expectations
+            k8s_resource: the current status/spec of the k8s resource
+
+        Returns:
+            True/False
```
RedbackThomson, line 313: Same as above, regarding the comment

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -176,23 +195,85 @@ def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceRe
...
+            if self.assert_extra_items_aws(aws_resource, key, value):
+                continue
+            assert value == aws_resource.get(key)
+
+    def get_aws_resource(self, resource_name):
+        """get aws resource from boto3
+        Override it for each resource type
+
+        Args:
+            k8s_resource: the current status/spec of the k8s resource
```
RedbackThomson, line 248: You renamed this parameter

RedbackThomson (commented):
I really only have minor nit picks left, otherwise I think this code is now excellent

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Added explanation.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Added.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Deleted unused arg

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Deleted unused arg

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Made them consistent.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Changed it.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -201,16 +282,59 @@ def assert_items(self, expectations: dict, state: dict) -> None:
...
+                continue
+            # assert any specific extra fields
+            if self.assert_extra_items_k8s(key, value, k8s_resource):
+                continue
+            assert value == k8s_resource.get(key)
+
+    def assert_extra_items_k8s(self, expected_name, expected_value, k8s_resource) -> bool:
+        """Asserts extra fields from k8s resource against fields from expectations
+        Override it as needed for custom verifications
+        Return true if the expected_name is asserted successfully
```
jljaco, line 306: I think there is room for improvement in this documentation on what the return value should be; it doesn't make sense to say that "expected_name is asserted successfully".  Can you describe more clearly what are the conditions for which this return value should be `True`?

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -201,16 +282,59 @@ def assert_items(self, expectations: dict, state: dict) -> None:
...
+            expected_value: expected value of a specific field to assert in k8s resource
+            k8s_resource: the current status/spec of the k8s resource
+
+        Returns:
+            True/False
+        """
+        return False
+
+    def assert_tags(self, expected_tags, latest_tags):
+        """Asserts tags from resource against tags from expectations
```
jljaco, line 320: I think you can rewrite this to more clearly state what must be true about `expected_tags` and `latest_tags` in order for this function to not violate an assertion

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Explained it with clearer words.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Added more words to explain it.

test/e2e/tests/test_scenarios.py
```diff
@@ -57,6 +121,11 @@ class UserHelper(helper.ResourceHelper):
     Overrides methods as required for custom resources.
     """
 
+    def get_aws_resource(self, resource_name):
+        users = self.mdb.describe_users(UserName=resource_name)
+        user = users['Users'][0]
+        return user
+
     def wait_for_delete(self, reference: k8s.CustomResourceReference):
         logging.debug(f"User - wait_for_delete()")
```
A-Hilaly, line 130: Just curious here, how are these function waiting for the resources to be deleted? are they overridden somewhere?

A-Hilaly (commented):
Good progress on this! :+1: 
I have one question

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +114,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
         sleep(self.DEFAULT_WAIT_SECS)
         self.wait_for_delete(reference)  # throws exception if wait fails
 
-    def assert_expectations(self, verb: str, input_data: dict, expectations: model.ExpectDict, reference: k8s.CustomResourceReference) -> None:
-        """Asserts custom resource reference inside Kubernetes cluster against the supplied expectations
+    def assert_k8s_resource(self, verb: str, input_data: dict, expectations_k8s: model.ExpectK8SDict, reference: k8s.CustomResourceReference) -> None:
+        """Compare the supplied expectations with custom resource reference inside Kubernetes cluster
 
-        :param verb: expectations after performing the verb (apply, patch, delete)
+        :param verb: compare expectations with k8s resource after performing the verb (apply, patch, delete)
```
RedbackThomson, line 120: nit: We aren't using this verb parameter in this method

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -110,21 +114,36 @@ def delete(self, reference: k8s.CustomResourceReference) -> None:
...
+        # status and spec assertions with resources from Kubernetes
+        self.assert_items_k8s(expectations_k8s.get("status"), resource.get("status"))
+        self.assert_items_k8s(expectations_k8s.get("spec"), resource.get("spec"))
+
+    def assert_aws_resource(self, verb: str, input_data: dict, expectations_aws: dict, reference: k8s.CustomResourceReference) -> None:
+        """Compare the supplied expectations with aws resource from api calls by boto3
 
-        # self._assert_state(expectations.get("spec"), resource)  # uncomment to support spec assertions
+        :param verb: compare expectations with aws resource after performing the verb (apply, patch, delete)
+        :param input_data: input data to verb
```
RedbackThomson, line 137: nit: We aren't using `verb` or `input_data` in this method

test/e2e/declarative_test_fwk/helper.py
RedbackThomson: I don't see them?

RedbackThomson (approved):
I think you might have forgotten to  push a change - you said you've updated the docs for some of the methods but I don't see them. Otherwise good to go!

test/e2e/tests/test_scenarios.py
kyriechen96: There is no operation for `wait_for_delete` in the controller. Does `runtime` call deletion again if the resource is not deleted? I remember I saw `reconciler` called `deleteResource ` again in the controller log after some default wait time?

kyriechen96:
> c183ce9

@RedbackThomson [c183ce9](https://github.com/aws-controllers-k8s/memorydb-controller/pull/65/commits/c183ce9876062a012abddb95c788d83df75fcfc1) this one?

test/e2e/tests/test_scenarios.py
A-Hilaly: That's why I asked, what's the point of having a `wait_for_delete` is none of the resources will be hang a little bit before being deleted. But it's a nit please ignore.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: I added above in the comment part not in returns part.
```
If resource has expected_name as a field, execute custom logic to validate that specific field and return true.
Otherwise, if resource doesn't have expected_name as a field, return false
```

test/e2e/declarative_test_fwk/helper.py
kyriechen96:
```
  def assert_extra_items_aws(self, expected_name, expected_value, aws_resource) -> bool:
        """Asserts specific fields that boto3 response against supplied expectation
        Override it as needed for validations of specific field
        If resource has expected_name as a field, execute custom logic to validate that specific field and return true
        Otherwise, if resource doesn't have expected_name as a field, return false
        Args:
            expected_name: expected name of a specific field to assert in aws resource
            expected_value: expected value of a specific field to assert in aws resource
            aws_resource: specific resource from aws memorydb
        Returns:
            True/False
        """
        return False
```

test/e2e/declarative_test_fwk/helper.py
kyriechen96: def assert_extra_items_aws(self, expected_name, expected_value, aws_resource) -> bool:
       """Asserts specific fields that boto3 response against supplied expectation
       Override it as needed for validations of specific field
       If resource has expected_name as a field, execute custom logic to validate that specific field and return true
       Otherwise, if resource doesn't have expected_name as a field, return false
       Args:
           expected_name: expected name of a specific field to assert in aws resource
           expected_value: expected value of a specific field to assert in aws resource
           aws_resource: specific resource from aws memorydb
       Returns:
           True/False
       """
       return False

RedbackThomson: Oh ok. Usually we include the comments about the return value on the same line as where we define its type. This is the "Google docstring format" which it seems like the original author of this framework was following. Here are the docs for examples on using it - https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Deleted them

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Deleted them.

test/e2e/declarative_test_fwk/helper.py
kyriechen96: Modified the words base on @jljaco comment https://github.com/aws-controllers-k8s/memorydb-controller/pull/65#discussion_r1122265736

kyriechen96: Modified the format!

test/e2e/tests/test_scenarios.py
kyriechen96: I can ask the original author about it.

A-Hilaly (approved):
:+1: Thank you for you patience. Good work on this!

nmvk (approved):

jljaco (approved):

PR #59, opened 2023-02-08
Integrate functions for tags from hooks.go files and add unit test for it

kyriechen96:
Issue #, if available:

Description of changes:
Integrate functions for tags from hooks.go files and add unit test for it.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #67, opened 2023-02-21
Update updateTags function by using new runtime function

kyriechen96:
Issue #, if available:
New method of comparing tags was added in runtime. Using this new method to reduce amount of custom logic code of updateTags.

Description of changes:
Use new method in runtime to replace custom logic of comparing two tags.


RedbackThomson (approved):
Approving on the basis that PR #59 will be rebased with these changes and tests will be added

PR #74, opened 2023-03-13
Fix incorrect err name

kyriechen96:
Issue #, if available:
Controller shuts down if calling `ACL` update while `state` of `ACL` is not `active`.

Description of changes:
One of error name doesn't match its return value. Correct it to match return value.


RedbackThomson (approved):
/lgtm
Nice easy fix. Thank you

PR #75, opened 2023-03-15
Support updating all parameters at once

kyriechen96:
Issue #, if available:
`update-parameter-group` API has a maximum limitation to parameters update. Only 20 parameters are allowed to update at once.

Description of changes:
Add `for` loop for calling `update-parameter-group` API to update up to 20 parameters each time.


pkg/resource/parameter_group/hooks.go
```diff
@@ -158,6 +159,141 @@ func (rm *resourceManager) resetAllParameters(
...
 
+// customUpdate overrides sdkUpdate by custom logic
+func (rm *resourceManager) customUpdate(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+	delta *ackcompare.Delta,
+) (updated *resource, err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.sdkUpdate")
```
RedbackThomson, line 170: Please update this to `customUpdate` instead of `rm.sdkUpdate`

pkg/resource/parameter_group/hooks.go
```diff
@@ -158,6 +159,141 @@ func (rm *resourceManager) resetAllParameters(
...
+		if ko != nil || err != nil {
+			return ko, err
+		}
+	}
+
+	if !delta.DifferentExcept("Spec.Tags") {
+		return desired, nil
+	}
+
+	input1, input2 := rm.updateRequestPayload(desired)
```
RedbackThomson, line 193: I don't think this will scale. Although there are only 39 parameters for now, there may be more later. Rather than update this code again, I would rather we write it with no limits on the maximum number of updates.

RedbackThomson (changes requested):

pkg/resource/parameter_group/hooks.go
kyriechen96: Modified.

pkg/resource/parameter_group/hooks.go
kyriechen96: Updated to a for loop.

pkg/resource/parameter_group/hooks.go
```diff
@@ -158,6 +159,119 @@ func (rm *resourceManager) resetAllParameters(
...
+		arn := ackv1alpha1.AWSResourceName(*resp.ParameterGroup.ARN)
+		ko.Status.ACKResourceMetadata.ARN = &arn
+	}
+	if resp.ParameterGroup.Description != nil {
+		ko.Spec.Description = resp.ParameterGroup.Description
+	} else {
+		ko.Spec.Description = nil
+	}
+	if resp.ParameterGroup.Family != nil {
+		ko.Spec.Family = resp.ParameterGroup.Family
```
RedbackThomson, line 224: This is going to throw an error if `resp` is nil. Also is `resp.ParameterGroup.Family` going to be the same for every request?

pkg/resource/parameter_group/hooks.go
```diff
@@ -158,6 +159,119 @@ func (rm *resourceManager) resetAllParameters(
...
+}
+
+// updateRequestPayload returns an array of SDK-specific struct for the HTTP request
+// payload of the Update API call for the resource
+// each element in return value has maximum 20 parameters
+func (rm *resourceManager) updateRequestPayload(
+	desired *resource,
+) ([]*svcsdk.UpdateParameterGroupInput) {
+	inputs := []*svcsdk.UpdateParameterGroupInput{}
+	parameterNameValues := lo.Chunk[*svcapitypes.ParameterNameValue](desired.ko.Spec.ParameterNameValues, 20)
```
RedbackThomson, line 250: If you could make `20` into a const at the top of this file, that will make it clearer why we have to chunk it

RedbackThomson: Also you can remove the `[*svcapitypes.ParameterNameValue]`. The compiler should be smart enough to know what type is returned, using inference.

pkg/resource/parameter_group/hooks.go
```diff
@@ -158,6 +159,119 @@ func (rm *resourceManager) resetAllParameters(
...
+
+	for _, iter := range parameterNameValues {
```
RedbackThomson, line 252: nit: This could be even simpler if you did `return lo.Chunk(desired.ko.Spec.ParameterNameValues, 20).Map(...)`. You could make this into a single statement

RedbackThomson (changes requested):

pkg/resource/parameter_group/hooks.go
kyriechen96: A `parameter group` resource only has one value to `Family` field. If these multiple requests are all for this specific resource, then `Family` field should always be the same value. 
I will add an empty `UpdateParameterGroupWithContext` call to make sure `resp` is not `nil`.

pkg/resource/parameter_group/hooks.go
kyriechen96: Add a constant for it!

pkg/resource/parameter_group/hooks.go
kyriechen96: I use `lo.Map` to replace the outer `for` loop.

pkg/resource/parameter_group/hooks.go
RedbackThomson: I'd rather we don't call `UpdateParameterGroupWithContext` unnecessarily. We could hit limits more quickly if we call it more than we need to. Instead, we can just check `if resp != nil {}` before we make any changes that require `resp`

kyriechen96: Modified.

pkg/resource/parameter_group/hooks.go
```diff
@@ -158,6 +163,121 @@ func (rm *resourceManager) resetAllParameters(
...
+
+// updateRequestPayload returns an array of SDK-specific struct for the HTTP request
+// payload of the Update API call for the resource
+// each element in return value has maximum 20 parameters
+func (rm *resourceManager) updateRequestPayload(
+	desired *resource,
+) []*svcsdk.UpdateParameterGroupInput {
+	inputs := []*svcsdk.UpdateParameterGroupInput{}
+	parameterNameValues := lo.Chunk(desired.ko.Spec.ParameterNameValues, maxNumberOfParametersUpdate)
+	inputs = lo.Map(parameterNameValues, func(parameters []*svcapitypes.ParameterNameValue, index int) *svcsdk.UpdateParameterGroupInput {
```
RedbackThomson, line 258: nit: We can just add `return` to the start of this line directly, then we can delete L256 and L278

pkg/resource/parameter_group/hooks.go
```diff
@@ -158,6 +163,121 @@ func (rm *resourceManager) resetAllParameters(
...
+	}
+
+	if !delta.DifferentExcept("Spec.Tags") {
+		return desired, nil
+	}
+
+	inputs := rm.updateRequestPayload(desired)
+
+	var resp *svcsdk.UpdateParameterGroupOutput
+	_ = resp
```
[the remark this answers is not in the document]
kyriechen96: Delete line 200

pkg/resource/parameter_group/hooks.go
kyriechen96: Modified

RedbackThomson (approved):
/lgtm

PR #77, opened 2023-03-20
Return an error message if snapshot is not ready for deleting

kyriechen96:
Issue #, if available:
If `snapshot` is not ready for deleting and is called by `sdkDelete`, it will have an error `InvalidSnapshotStateFault`.

Description of changes:
Check whether `snapshot` is in the `status` to be deleted. If not, then output an error message.

### DIFF
--- a/pkg/resource/snapshot/utility.go
+++ b/pkg/resource/snapshot/utility.go
@@ -2,6 +2,7 @@ package snapshot
 
 import (
 	"errors"
+	"fmt"
 	svcapitypes "github.com/aws-controllers-k8s/memorydb-controller/apis/v1alpha1"
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
@@ -11,12 +12,18 @@ import (
 
 var (
 	condMsgCurrentlyDeleting string = "snapshot currently being deleted"
+	availableStatus          string = "available"
 	deleteStatus             string = "deleting"
+	failedStatus             string = "failed"
 )
 
 var (
 	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
 		errors.New("delete is in progress"),
+		ackrequeue.DefaultRequeueAfterDuration,
+	)
+	requeueWaitSnapshotIsReadyForDeleting = ackrequeue.NeededAfter(
+		fmt.Errorf("snapshot is not ready for deletion - must be in either %q or %q state", availableStatus, failedStatus),
 		ackrequeue.DefaultRequeueAfterDuration,
 	)
 )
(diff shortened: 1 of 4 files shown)


A-Hilaly (approved):
Nice! 👍

pkg/resource/snapshot/utility.go
```diff
@@ -30,6 +37,15 @@ func isDeleting(r *resource) bool {
 	return status == deleteStatus
 }
 
+// isNotReadyForDeleting returns true if supplied cluster resource state is 'copying', 'creating', or 'restoring'
+func isNotReadyForDeleting(r *resource) bool {
+	if r == nil || r.ko.Status.Status == nil {
+		return false
+	}
+	status := *r.ko.Status.Status
+	return status == copyingStatus || status == creatingStatus || status == restoringStatus
```
RedbackThomson, line 46: I would do this the other way around. Check if `status != available && status != failed` - since that's what the service will check anyway

pkg/resource/snapshot/utility.go
```diff
@@ -12,13 +12,20 @@ import (
...
+	restoringStatus          string = "restoring"
 )
 
 var (
 	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
 		errors.New("delete is in progress"),
 		ackrequeue.DefaultRequeueAfterDuration,
 	)
+	requeueWaitSnapshotIsReadyForDeleting = ackrequeue.NeededAfter(
+		errors.New("snapshot is not ready for deleting. snapshot must be in either available or failed state in order to be deleted"),
```
RedbackThomson, line 26: nit: use string interpolation:
```suggestion
fmt.Errorf("snapshot is not ready for deletion - must be in either %q or %q state", availableStatus, failedStatus),
```

RedbackThomson (changes requested):

pkg/resource/snapshot/utility.go
kyriechen96: There is an issue here. `InvalidSnapshotStateFault` shows `state` of `snapshot` has to be `available` or `failed`. However, when I check `describe-snapshot` API, `status` of `snapshot` doesn't have `failed` as an option. Hence, I'm not sure if there is `failed` `status` for `snapshot`. https://docs.aws.amazon.com/cli/latest/reference/memorydb/describe-snapshots.html
Also, if we do the another way, then it might should be `status != available && status != failed && status != deleting`

kyriechen96:
```Status -> (string)

The status of the snapshot. Valid values: creating | available | restoring | copying | deleting.
```
`available` is there.

kyriechen96: I'm checking with my team for it now.

pkg/resource/snapshot/utility.go
kyriechen96: Modified.

pkg/resource/snapshot/utility.go
kyriechen96: It has "failed" as `status` for `snapshot`.

RedbackThomson (approved):
/lgtm

